PR #12, opened 2024-04-29
resolve broken links and frontmatter structure

LordGhostX:
This PR resolves the broken links and frontmatter in the specs markdown files per https://github.com/waku-org/waku.org/pull/76


jm-clius (approved):
Thanks for this!

PR #14, opened 2024-05-17
feat: message references with DAG

kaichaosun:
This spec proposes a way to reference messages from group members when sending a new message. In this way, the messages of the group compose a Directed Acyclic Graph (aka. DAG) effectively. By looking into the tips of the graph, the member in the group can get the missing messages from store node with the reference specified in the tip message. Each member also send a checkpoint message periodically.

### DIFF
--- a/standards/application/dag-ack.md
+++ b/standards/application/dag-ack.md
@@ -27,7 +27,7 @@ The message which is not input by user, but create and send automatically in app
 It's a type of control message from a group member, which only contains references. Such message should not be persistent in any nodes, instead it's only used to find out if the referenced message is missing or not by other group members.
 
 **Request message**
-It's another type of control message for asking a message by its reference id. It can be persistent in store and sender node but not others. Group members will only fulfill the request message if store node doesn't have it. The sender node regularly check if pending request message gets fulfilled or not. To avoid the same request message flood by many members in one group, the sender node should not send new messages if there already exists such request message in the store node (using payload based query for such check).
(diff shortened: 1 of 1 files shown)


standards/application/dag-ack.md
```diff
@@ -0,0 +1,155 @@
...
+
+picked_members = weighted_choose(active_group, inactive_group, prob_active, k)
+
+print("Picked members:", picked_members)
+```
+
+**Note**: It's recommended to always include a reference to your previous message, and two or three references to other group members. If the group is small (less than 5 members), one reference to other members is acceptable. If the group is large and the length of message payload is not a limitation, following curve can be considered,
+
+```
+f = a * log(x)
```
jm-clius, line 120: Could you explain a bit more how this curve would work for very large groups? For Community protocol our eventual aim is to support Communities of up to 10K members. Presumably this would create a significant bandwidth and processing overhead?

standards/application/dag-ack.md
```diff
@@ -0,0 +1,155 @@
...
+concat(message_hash_1, message_hash_2, message_hash_3)
+```
+
+### Checkpoint Message
+
+The payload for `Checkpoint Message` is similar to Tip Message, but it only contains references, no other attributes.
+
+### Request Message
+
+The payload for `Request Message` is defiled as following,
```
jm-clius, line 73: Just to be double sure I understand - say Alice believes she has sent messages `a1`..`a5`, each including a reference to the previous message. However, messages `a1`..`a4` actually failed to send due to a connectivity issue. The first successfully broadcast message is `a5`. Everyone in the group will see a reference to `a4`, which is missing, request this message and then recursively request all messages back in time to `a1`? Of course, we can try to avoid all recipients responding with a `request` at once by having them backoff randomly and check the store node in case another group member has requested the missing messages, but presumably there may still be a fairly large spike in activity for very large groups?

jm-clius (commented):
Still considering some parts of this, but have added two more questions. :)

PR #18, opened 2024-05-27
Add "req-res protocol reliability" spec

weboko:
This PR adds initial version of `Reliability for Request-Response` protocols.

In here you can find summary of either ongoing or already finished efforts from `status-go` and `js-waku`.


informational/req-res-reliability.md
```diff
@@ -0,0 +1,62 @@
...
+
+- In case a node goes offline while having an active subscription - it is important to do ping again right after node appears online. In case ping fails - re-subscribe request should be fired to a new peer.
+
+- While registering Filter subscriptions - it is advised to batch requests for multiple content topics into one in order to reduce amount of queries sent to a node. 
+
+- During creation of a new subscription it can be benefitial to use only new peers to which no subscriptions yet present and not use peers with which Filter already failed.
+
+## Security/Privacy Considerations
+
+None of the mentioned recommendations incur privacy or security tradeoffs and in some cases increase k-anonymity (e.g having unique peers for Filter subscriptions).
```
jm-clius, line 58: Not sure I follow what you mean with "having unique peers for Filter subscriptions"? If I subscribe my filter in multiple service nodes, how does that increase my anonymity?

jm-clius (changes requested):
All in all a great start!

I've added some comments below. In general:
- we have to define from which POV we're defining reliability and be explicit throughout - in this case the POV of the _requester_ within req-resp protocols. We can use "client", "requester" or both, as long as we define these terms in the beginning and use them consistently throughout.
- we should be more clear about tradeoffs affecting decisions such as number of connections to service nodes, failure thresholds, etc.
- one section that's missing might be a possible interaction between lightpush and filter. For example, if I have an active filter subscription to a service node different from the service node I use for `lightpush`, couldn't I simply verify that I receive my own published messages thereby reducing need for redundant publish? There are some assumptions of course, such as overlapping subscription, but I think this is a valid reliability strategy.

informational/req-res-reliability.md
jm-clius: This seems to me to more closely address concerns re censorship resistance. My anonymity is equally vulnerable in all three service nodes, although I may indeed spread the risk of a specific message being deanonymised by using different service nodes as entry point. However, the spec mentions Filter specifically, which I'm a bit more confused by. 😅

informational/req-res-reliability.md
```diff
@@ -0,0 +1,62 @@
...
+
+- While sending message with Light Push - it is advised to use more than 1 peer in order to increase changes of delivering message.
+
+- If sending message is failed to all of the peers - node should try to re-send message after some interval and continue doing so while OK response is received. 
+
+### Filter
+
+- To decrease chances of missing messages a node can initiate more than one subscription through Filter protocol to the same content topic and filter out duplicates. This will increase bandwidth consumption and would depend on the information exchanged under content topic in use.
+
+- In case a node goes offline while having an active subscription - it is important to do ping again right after node appears online. In case ping fails - re-subscribe request should be fired to a new peer.
```
chaitanyaprem, line 50: Interesting approach, but i have seen that it is practically not possible because detecting a node went offline takes time ~20-30seconds in which time the peer ping might end up failing and causing subscriptions to get removed.  
Also, note that Filter-ping doesn't have enough information to know which are the subscriptions currently active with a peer..it only indicates that there is a subscription with it. 
Currently in status-go/go-waku the way it is implemented is as soon as offline status is identified all subscriptions are cleared off. When node comes back online new subscriptions are created. 

At the same time, i am curious to know if we have taken the apporach mentioned here in js-waku what was the efficacy with node connectivity and did we not run into any issues?

informational/req-res-reliability.md
```diff
@@ -0,0 +1,62 @@
...
+## Motivation
+
+Descriptions of mentioned protocols do not define some of the real world use cases that are oftenly observed in unreliable network environment. Such use cases can be: recovery from offline state, decrease rate of missed messages, increase probability of messages being broadcasted within the network.
+
+## Suggestions
+
+### Node health
+
+As a useful metric to define and implement for determining quality of provided service by a node:
+- unhealthy - no peer connections are available regardless of protocol;
```
chaitanyaprem, line 24: I am rethinking about health of a lightClient, maybe peer connections is not the right way. 
Rather whether we have subscriptions for all filters active and if we have peers that we can send messages via lightpush. I will be working on it this week as to how do we define health in case of lightClient. Maybe we can update this based on the work that goes into it.

informational/req-res-reliability.md
```diff
@@ -0,0 +1,62 @@
...
+- minimally healthy:
+  - Relay has less than 4 peers connected;
+  - Filter and LightPush has one per each peer connection available;
+- sufficiently healthy:
+  - Relay has minimum 4 peers connected;
+  - more than 1 connection in Filter and at least 2 connections available in LightPush;
+
+### Peers and connection management
+
+- Each protocols should retain a pool of reliable peers. In case a protocol failed to use any peer more than once - connection to it should be dropped and new peer should be added to the pool instead.
```
chaitanyaprem, line 34: Rather than calling it protocol, we should indicate that each client should retain a pool of peers for Filter and lightPush. 

Also i am wondering if we have an understanding of what i means by using a peer. We are looking at a peer-to-peer network where serviceNodes availability is not gauranteed unless they are fleet nodes. e.g: status-desktop instances may not be always running. 
There could be different scenarios why connectivity to a specific peer fails at specified time such as rate limits reached or ip-colocation-limit hit . That doesn't necessarily mean it is a bad peer. Hopefully we are not referring to descoring the peer.

informational/req-res-reliability.md
```diff
@@ -0,0 +1,62 @@
...
+  - Filter and LightPush has one per each peer connection available;
+- sufficiently healthy:
+  - Relay has minimum 4 peers connected;
+  - more than 1 connection in Filter and at least 2 connections available in LightPush;
+
+### Peers and connection management
+
+- Each protocols should retain a pool of reliable peers. In case a protocol failed to use any peer more than once - connection to it should be dropped and new peer should be added to the pool instead. 
+
+- During discovery of new peers it is better to filter unwonted out based on ENR / multiaddress. For example in some cases `circuit-relay` addresses are not needed when we try to find and connect to peers directly.
```
chaitanyaprem, line 36: to filter unwonted out -> to filter out

informational/req-res-reliability.md
```diff
@@ -0,0 +1,62 @@
...
+
+- When peer is discovered second time, we need to be sure to keep connection information up to date in Peer Store.
```
chaitanyaprem, line 38: Better be specific when we talk about what information we keep like ENR, peer's multiaddress and its capabilities etc.

informational/req-res-reliability.md
```diff
@@ -0,0 +1,62 @@
...
+
+### Light Push 
+
+- While sending message with Light Push - it is advised to use more than 1 peer in order to increase changes of delivering message.
```
chaitanyaprem, line 42: Not sure if this is recommended still, in status case only 1 lightpush node is being used and along with regular storev3 queries to confirm message deliver, we don't need to use more than 1 lightpush node.
Do we have any data in case of js-waku indicating issues with using only 1 lightpush peer and hence switching to more than 1?

Also with changes to lightpush protocol coming in, we can just rely on serviceNode response to determine if we want to use a second lightpush peer or not.

informational/req-res-reliability.md
```diff
@@ -0,0 +1,62 @@
...
+
+- If sending message is failed to all of the peers - node should try to re-send message after some interval and continue doing so while OK response is received.
```
chaitanyaprem, line 44: Should this be unlimited retry? Maybe not be a good idea, better to always have a limit on number of times we send.

informational/req-res-reliability.md
```diff
@@ -0,0 +1,62 @@
...
+category: Best Current Practice
+tags: [informational]
+editor: Oleksandr Kozlov <oleksandr@status.im>
+contributors:
+- Prem Chaitanya Prathi <prem@status.im>
+- Danish Arora <danish@status.im>
+---
+
+## Abstract
+This RFC describes set of instructions used across different [WAKU2](https://github.com/vacp2p/rfc-index/blob/7b443c1aab627894e3f22f5adfbb93f4c4eac4f6/waku/standards/core/10/waku2.md?plain=1#L3) implementations for improved reliability in request-response protocols such as [WAKU2-LIGHTPUSH](https://github.com/vacp2p/rfc-index/blob/7b443c1aab627894e3f22f5adfbb93f4c4eac4f6/waku/standards/core/19/lightpush.md?plain=1#L3C11-L3C26) and [WAKU2-FILTER](https://github.com/vacp2p/rfc-index/blob/7b443c1aab627894e3f22f5adfbb93f4c4eac4f6/waku/standards/core/12/filter.md?plain=1#L3).
```
chaitanyaprem, line 13: Maybe good to somewhere mention that Lightpush is used to send messages and Filter is used to receive them.

informational/req-res-reliability.md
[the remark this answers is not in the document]
chaitanyaprem: Filtering is done based on criteria such as shards, capabilities (i.e if the peer supports lightpush and Filter). Better to mention specific filtering that is done.

informational/req-res-reliability.md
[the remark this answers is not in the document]
chaitanyaprem: Yes, we need to indicate that regular filter pings helps in identifying loss of subscription which in-turn can be used as a trigger to open a subscription to a different peer.

informational/req-res-reliability.md
```diff
@@ -0,0 +1,62 @@
...
+- In case a node goes offline while having an active subscription - it is important to do ping again right after node appears online. In case ping fails - re-subscribe request should be fired to a new peer.
+
+- While registering Filter subscriptions - it is advised to batch requests for multiple content topics into one in order to reduce amount of queries sent to a node. 
+
+- During creation of a new subscription it can be benefitial to use only new peers to which no subscriptions yet present and not use peers with which Filter already failed.
+
+## Security/Privacy Considerations
+
+None of the mentioned recommendations incur privacy or security tradeoffs and in some cases increase k-anonymity (e.g having unique peers for Filter subscriptions).
+
```
chaitanyaprem, line 59: I don't see a section on regular store queries. But maybe that is already covered as part of relay reliability spec and we can just reference that section here indicating how it is used to identify missing messages (both sent and supposed to receive) and how fetch/resend is done to fill the gaps.

informational/req-res-reliability.md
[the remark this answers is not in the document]
weboko: 💯

informational/req-res-reliability.md
```diff
@@ -0,0 +1,62 @@
...
+- Prem Chaitanya Prathi <prem@status.im>
+- Danish Arora <danish@status.im>
+---
+
+## Abstract
+This RFC describes set of instructions used across different [WAKU2](https://github.com/vacp2p/rfc-index/blob/7b443c1aab627894e3f22f5adfbb93f4c4eac4f6/waku/standards/core/10/waku2.md?plain=1#L3) implementations for improved reliability in request-response protocols such as [WAKU2-LIGHTPUSH](https://github.com/vacp2p/rfc-index/blob/7b443c1aab627894e3f22f5adfbb93f4c4eac4f6/waku/standards/core/19/lightpush.md?plain=1#L3C11-L3C26) and [WAKU2-FILTER](https://github.com/vacp2p/rfc-index/blob/7b443c1aab627894e3f22f5adfbb93f4c4eac4f6/waku/standards/core/12/filter.md?plain=1#L3).
+
+## Motivation
+
+Descriptions of mentioned protocols do not define some of the real world use cases that are oftenly observed in unreliable network environment. Such use cases can be: recovery from offline state, decrease rate of missed messages, increase probability of messages being broadcasted within the network.
```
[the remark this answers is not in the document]
weboko: sounds better

informational/req-res-reliability.md
weboko: > detecting a node went offline takes time ~20-30seconds

that is not the case for browser, we can identify that it is offline almost instantly 

> When node comes back online new subscriptions are created.

that is interesting and this is what can kinda happen, as the described approach in the worst case will re-initiate all the subscriptions

> did we not run into any issues

it is still experimental but on my local repros I observe the issue when subscription is dropped by a service node, that is happening somewhere after 10-15 mins.
Considering that we might just not do a ping if a node was offline for more than 10 minutes, but the problem here is it is not controllable behavior, meaning there is no CLI flag specified and it is not described in Filter RFC so it is totally up to an implementation.

informational/req-res-reliability.md
weboko: let's work together on that this week then, if it will take longer we can always follow up with a PR

informational/req-res-reliability.md
weboko: agree, added

informational/req-res-reliability.md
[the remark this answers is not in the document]
weboko: makes sense, will refactor sections one by one while making requested changes

informational/req-res-reliability.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Suggestions
+
+### Node health
+
+As a useful metric to define and implement for determining quality of provided service by a node:
```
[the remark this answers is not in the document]
weboko: updated Abstract and Motivation sections, added Definitions sub section to clarify `service node`/`light node`
https://github.com/waku-org/specs/pull/18/commits/ba103d2808b113d8bd33f274c55688bc56cab021

informational/req-res-reliability.md
```diff
@@ -0,0 +1,62 @@
...
+- unhealthy - no peer connections are available regardless of protocol;
+- minimally healthy:
+  - Relay has less than 4 peers connected;
```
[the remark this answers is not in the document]
weboko: it was more of a copy-paste, agree relay is not needed here and will be addressed by another RFC

informational/req-res-reliability.md
```diff
@@ -0,0 +1,62 @@
...
+  - Filter and LightPush has one per each peer connection available;
+- sufficiently healthy:
+  - Relay has minimum 4 peers connected;
+  - more than 1 connection in Filter and at least 2 connections available in LightPush;
```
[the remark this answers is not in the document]
weboko: I agree, updated section. A bit too repetitive language but I think for starters good enough. We will have a bit more discussions around this section though https://github.com/waku-org/specs/pull/18#discussion_r1660824855

https://github.com/waku-org/specs/pull/18/commits/c23dfe95f5987a07fec74e807dad506ead9a9ae1

informational/req-res-reliability.md
```diff
@@ -0,0 +1,62 @@
...
+## Suggestions
+
+### Node health
+
+As a useful metric to define and implement for determining quality of provided service by a node:
+- unhealthy - no peer connections are available regardless of protocol;
+- minimally healthy:
+  - Relay has less than 4 peers connected;
+  - Filter and LightPush has one per each peer connection available;
+- sufficiently healthy:
```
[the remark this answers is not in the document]
weboko: improved definitions and provided context in there https://github.com/waku-org/specs/pull/18/commits/ef1c08a15fbd921ab916ca0e1b776c3442a14778

informational/req-res-reliability.md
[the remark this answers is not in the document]
weboko: addressed here - https://github.com/waku-org/specs/pull/18/commits/594a1ecf13ab5665e749b9adaab168fa77020a25

> We can leave precise definition of "bad service" to the individual protocol specs.

do you see it being a follow up to this PR or, perhaps, I can add it as separate section that defines more precisely `fails to serve protocol request` (as I defined it rn)

> We can then continue to say that we recommend a very strict approach of removing service nodes after even 1 failure. (Do we really recommend this, though? If we fail to connect to a service node, should we completely remove it from the pool even if it might just be a temporary connection failure?)

I think this change I did before we made actual code change so that's why it was `more than once` and I am changing it to `fails to serve protocol request from a light node 3 times`. I am not sure if it is best framing though

informational/req-res-reliability.md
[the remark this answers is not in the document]
weboko: added shorter explanation but also keeping prev sentence to give additional context. wdyt @jm-clius ?
https://github.com/waku-org/specs/pull/18/commits/6a149261b6e758031230d6eb7879c76895aea871

informational/req-res-reliability.md
weboko: removed in favor of https://github.com/waku-org/specs/pull/18/commits/1f433b90eaa922b88228e210a27c51046898319d

informational/req-res-reliability.md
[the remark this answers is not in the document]
weboko: now it should be more explicit

informational/req-res-reliability.md
```diff
@@ -0,0 +1,89 @@
...
+These preferences might include:
+- [Libp2p multiadresses](https://github.com/libp2p/specs/blob/master/addressing/README.md) of a service node;
+- Waku or libp2p protocols that a service node implements;
+- Wakus shards that a service node is part of;
+
+More details about discovery can be found at [WAKU2 Discovery domain](https://github.com/vacp2p/rfc-index/blob/7b443c1aab627894e3f22f5adfbb93f4c4eac4f6/waku/standards/core/10/waku2.md#discovery-domain) or [RELAY-SHARDING Discovery](https://github.com/waku-org/specs/blob/master/standards/core/relay-sharding.md#discovery).
+
+Examples of filtering:
+- When [Circuit V2](https://github.com/libp2p/specs/blob/master/relay/circuit-v2.md) multiaddresses discovered by a light node - it might avoid connecting such service nodes and wait for service nodes that can be connected directly;
+- When light node discovers service nodes that implement needed Waku protocols - it should prioritize those that implement most recent version of protocol;
```
fryorcraken, line 59: At this point in time, we should aim for single version protocol support in light clients IMO.

We can look into multiple version support as the network grows and we have less control on it.

informational/req-res-reliability.md
```diff
@@ -0,0 +1,89 @@
...
+- Light node must connect only to those service nodes that participate in needed shard;
```
fryorcraken, line 60: yes, concept of "strict"

informational/req-res-reliability.md
```diff
@@ -0,0 +1,89 @@
...
+
+#### Continuous discovery
+Light node must keep information about service nodes up to date.
```
fryorcraken, line 63:
```suggestion
Light nodes must keep information about service nodes up to date.
```

informational/req-res-reliability.md
```diff
@@ -0,0 +1,89 @@
...
+For example when a service node is discovered second time,
+we need to be sure to keep connection information up to date in Peer Store.
+
+### Light Push
+
+- While sending message with Light Push - it is advised to use more than 1 peer in order to increase chances of delivering message.
+
+- If sending message is failed to all of the peers - node should try to re-send message after some interval and continue doing so until OK response is received.
```
fryorcraken, line 71: I think we should take the opportunity to define application concern (user of waku library) vs waku concern (reliablity in waku).

How much should waku res-send? At which point should it be handed over the applicaiton and by extension the end user?

One proposal was a temporal one: Waku owns the message to be sent for ~15s (under the 20s grace period) and attempt during this period to send the message and get peer-to-peer confirmation.

Also, we need to define "sending message failed" similarly to my previous comment: are we saying no response/error response from service node?
Or are we saying we were not able (via store or filter) to confirm the message was handed over to the network?

informational/req-res-reliability.md
fryorcraken: to that, I would define how to use store to recover.
Probably something like that (terminology can be improved):

- timeout: the underlying protocol timeout. 30s for TCP, 1s/0s for websocket ("instant"). This caps how fast one can detect a disconnection
- interval: time between filter pings
- t0: successful ping
- t1 (t0+interval): failed ping
- t2: successful ping

At t2: 
- recreate subscription
- trigger store v3 query to get msg id of missed messages.
   -  either use last received message with timestamp < t0 as cursor (likely best practice)
   - or do a time based query: start: t0, end: t2

informational/req-res-reliability.md
```diff
@@ -0,0 +1,62 @@
...
+
+- If sending message is failed to all of the peers - node should try to re-send message after some interval and continue doing so while OK response is received. 
+
+### Filter
+
+- To decrease chances of missing messages a node can initiate more than one subscription through Filter protocol to the same content topic and filter out duplicates. This will increase bandwidth consumption and would depend on the information exchanged under content topic in use.
+
+- In case a node goes offline while having an active subscription - it is important to do ping again right after node appears online. In case ping fails - re-subscribe request should be fired to a new peer.
+
+- While registering Filter subscriptions - it is advised to batch requests for multiple content topics into one in order to reduce amount of queries sent to a node.
```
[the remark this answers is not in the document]
fryorcraken: > Does this belong part of the reliability specifications? Seems more like part of the Filter specification

I think it;'s fine to discuss this matter here so we can have a consolidated discussion and then move it to filter spec when we want to merge.

informational/req-res-reliability.md
```diff
@@ -0,0 +1,89 @@
...
+
+### Filter
+
+- To decrease chances of missing messages a node can initiate more than one subscription through Filter protocol to the same content topic and filter out duplicates. This will increase bandwidth consumption and would depend on the information exchanged under content topic in use.
+
+- In case a node goes offline while having an active subscription - it is important to do ping again right after node appears online. In case ping fails - re-subscribe request should be fired to a new peer.
+
+- While registering Filter subscriptions - it is advised to batch requests for multiple content topics into one in order to reduce amount of queries sent to a node. 
+
+- During creation of a new subscription it can be beneficial to use only new peers to which no subscriptions yet present and not use peers with which Filter already failed.
```
fryorcraken, line 81: I think overall we need to have some sort of local ordering (scoring?) of peers to decide which peers should be used for x based on various criteria

informational/req-res-reliability.md
chaitanyaprem: I am curious to know how long does it take for websocket to detect timeout, because iirc websockets use TCP as underlying transport same as what is used in other waku implementations as well.
One possibility could be that websocket is having some sort of a very regular heartbeat which makes it detect quickly.

chaitanyaprem: > it is still experimental but on my local repros I observe the issue when subscription is dropped by a service node, that is happening somewhere after 10-15 mins.

The filter timeout has been set in nwaku to be 5 minutes, so it should get dropped after 5 minutes +/- some buffer time.

informational/req-res-reliability.md
chaitanyaprem: Sure, i have written down the approach for lightClients topic health [here](https://github.com/waku-org/go-waku/issues/1114#issuecomment-2201994802)

informational/req-res-reliability.md
weboko: added `Retry on failure` and `Retry missing messages` sub sections

informational/req-res-reliability.md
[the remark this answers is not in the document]
weboko: added link to swap of unreliable peer logic

weboko: true but our assumption would be in this case that if peer fails - it will be replaced with another one so we can continue with retry assuming a not failing peer will be hit eventually

please, check updated version of this section and if you comment holds true - let's discuss it agan

resolving for now

informational/req-res-reliability.md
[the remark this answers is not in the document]
weboko: both are super valid points - added `Regular pings` subsection

informational/req-res-reliability.md
```diff
@@ -0,0 +1,62 @@
...
+
+### Filter
+
+- To decrease chances of missing messages a node can initiate more than one subscription through Filter protocol to the same content topic and filter out duplicates. This will increase bandwidth consumption and would depend on the information exchanged under content topic in use.
+
+- In case a node goes offline while having an active subscription - it is important to do ping again right after node appears online. In case ping fails - re-subscribe request should be fired to a new peer.
+
+- While registering Filter subscriptions - it is advised to batch requests for multiple content topics into one in order to reduce amount of queries sent to a node. 
+
+- During creation of a new subscription it can be benefitial to use only new peers to which no subscriptions yet present and not use peers with which Filter already failed.
```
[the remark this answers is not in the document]
weboko: true, removed

informational/req-res-reliability.md
```diff
@@ -0,0 +1,89 @@
...
+category: Best Current Practice
+tags: [informational]
+editor: Oleksandr Kozlov <oleksandr@status.im>
+contributors:
+- Prem Chaitanya Prathi <prem@status.im>
+- Danish Arora <danish@status.im>
+---
+
+## Abstract
+This RFC describes set of instructions used across different [WAKU2](https://github.com/vacp2p/rfc-index/blob/7b443c1aab627894e3f22f5adfbb93f4c4eac4f6/waku/standards/core/10/waku2.md) implementations for improved reliability during usage of request-response protocols by a light node:
```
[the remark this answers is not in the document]
weboko: `waku2.md` does not exist in this repo so I am using persistent link to vacp2p repo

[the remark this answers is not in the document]
weboko: let's continue discussion in https://github.com/waku-org/specs/pull/18#discussion_r1663371253

informational/req-res-reliability.md
```diff
@@ -0,0 +1,89 @@
...
+- [WAKU2-LIGHTPUSH](https://github.com/vacp2p/rfc-index/blob/7b443c1aab627894e3f22f5adfbb93f4c4eac4f6/waku/standards/core/19/lightpush.md) - is used for sending messages;
```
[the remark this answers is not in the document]
weboko: as I mentioned in https://github.com/waku-org/specs/pull/18#discussion_r1671292130 - some files do not exist in this repo 

do we plan on moving them at any point? cc @jm-clius 

until it is done - I see the only way is using persistent GitHub links to `vacp2p` repo

informational/req-res-reliability.md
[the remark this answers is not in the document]
weboko: agree with @danisharora099 here - [Filter](https://github.com/vacp2p/rfc-index/blob/main/waku/standards/core/12/filter.md#filter-subscribe-request) has it covered

informational/req-res-reliability.md
weboko: added definition of failure for each protocol 

`descoring` is out of scope and we were considering something like that for js-waku in https://github.com/waku-org/js-waku/issues/2017

informational/req-res-reliability.md
weboko: specified enr and multiaddress, should be enough for now

informational/req-res-reliability.md
weboko: added `Retry missing messages` section that covers regular store queries and approach described in here https://github.com/waku-org/specs/pull/18#pullrequestreview-2144776176

informational/req-res-reliability.md
weboko: yeah, this is kinda [scoring](https://github.com/waku-org/js-waku/issues/2017) or some kind of pre arranged exchange of capabilities of a service node

we don't have any of these so maybe let's initiate discussion if we really want to have such a thing?

informational/req-res-reliability.md
```diff
@@ -0,0 +1,89 @@
...
+  - LightPush protocol has one service node connection;
+- sufficiently healthy:
+  - Filter has at least 2 connections available to service nodes;
+  - LightPush has at least 2 connections available to service nodes;
+
+### Peers and connection management
+
+#### Pool of reliable service nodes
+Light node should maintain a pool of reliable service nodes for each protocol.
+In case service node fails to serve protocol request from a light node 3 times - light node should drop connection to it and a new service node should be connected and added to the pool instead.
```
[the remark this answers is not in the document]
weboko: > 3 times seems high.

it can be, so with default `ping` interval it can be 3 minutes, and for `lightpush` it can be whatever it will take in order for user to send 3 requests 

adding better definition for light push and filter specifically

covered `a, b` as part of definition of failure
as for `c` I am not sure, I would consider it part of another strategy, will add a note to `Retry missing messages` subsection

 > If all remote nodes, then "dropping connection to it and using a new service node" is not really useful

I guess application should identify it is offline and/or their shard is not functioning

> If a single remote node is unreachable, then trying 3 times seems like a waste of time.

to me it seems as default assumption we do - service node fails 
but there are different kinds of failures: for Filter it is straightforward, probably, but for LightPush it might be different error that happened - I added a note saying that error code should be considered to make a decision to retry or drop.

informational/req-res-reliability.md
```diff
@@ -0,0 +1,89 @@
...
+
+#### Seection of discovered service nodes
+During discovery light node should filter out service nodes based on preferences before establishing connection.
+These preferences might include:
+- [Libp2p multiadresses](https://github.com/libp2p/specs/blob/master/addressing/README.md) of a service node;
+- Waku or libp2p protocols that a service node implements;
+- Wakus shards that a service node is part of;
```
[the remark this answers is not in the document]
weboko: added for strict

for preference - we need to somehow expose this information to be available, ideally, before connection is established, for some it is there already but - I believe it is tricky for others like latency and reputation.

also, this is useful to have for incentivisation, but then there should be some guarantee (validators?)
sound like a new feature :)

informational/req-res-reliability.md
weboko: I think it is still worth mentioning as it might be not only version but needed protocol depending on the strategies used: Store + LightPush or Filter + LightPush or all

informational/req-res-reliability.md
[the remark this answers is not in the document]
weboko: > 3 times seems high.

I am lowering it to 2 times and and setting in code 30 seconds for recurring pings
wdyt @danisharora099 ?

informational/req-res-reliability.md
weboko: added more definitions, this should be resolved

informational/req-res-reliability.md
weboko: @fryorcraken I would drop it out of this version of RFC
https://github.com/waku-org/specs/pull/18#discussion_r1663407256

this sounds like a more complex feature that should be useful but we don't have any precedent

what if I or someone from the team implements it in js-waku as part of improving Filter and after it we can add it to this RFC?

informational/req-res-reliability.md
```diff
@@ -0,0 +1,139 @@
...
+In case light node does not receive OK response or it times out 2 times - such service node should be replaced as part of maintenance of [pool of reliable service nodes](./req-res-reliability.md#pool-of-reliable-service-nodes).
+Right after such replace light node must create new subscription to newly connected service node as described in [Filter specification](https://github.com/vacp2p/rfc-index/blob/7b443c1aab627894e3f22f5adfbb93f4c4eac4f6/waku/standards/core/12/filter.md).
+
+#### Redundant subscriptions for message loss mitigation
+To mitigate possibility of messages not being delivered by a service node - we advice to consider using multiple Filter subscriptions. 
+Light node can initiate two subscriptions to the same content topic but to different service nodes. 
+While receiving messages through two subscriptions - duplicates must be dropped by using [deterministic message hashing](https://github.com/vacp2p/rfc-index/blob/main/waku/standards/core/14/message.md#deterministic-message-hashing).
+Note that such approach increase bandwidth consumption proportionally to amount of extra subscriptions established and should be used with caution. 
+
```
jm-clius, line 127: Another possible mechanism we should describe is regular "refresh" of subscriptions to ensure active subscriptions are still fully synchronised. Think of it as a much more occasional, more expensive "ping". Described here: https://github.com/vacp2p/rfc-index/blob/7b443c1aab627894e3f22f5adfbb93f4c4eac4f6/waku/standards/core/12/filter.md#subscribe
We do not have to recommend using this mechanism, as it's quite expensive and I'm not sure we'll gain much by implementing it. However, it is a mechanism affecting reliability and I think it's useful to have the entire toolbox briefly described.

jm-clius (approved):
Thank you! This is a good summary of available mechanisms for reliability and tradeoffs at play. I have added mostly minor comments, but this can be merged as a raw specification.

One thing that is still missing is that the protocol part of a spec should preferably be phrased the keywords MUST, SHOULD, RECOMMEND, etc. as described here: https://github.com/waku-org/specs/blob/master/template.md#wire-format-specification--syntax

I think this can be done in a follow-up or whenever this specification is revised.

informational/req-res-reliability.md
```diff
@@ -0,0 +1,139 @@
...
+#### Offline recoverability
+Network state should be monitored by light node and in case it goes offline - [regular pings](./req-res-reliability.md#regular-pings) must be stopped. 
+When network connection returns light node should initiate [Filter ping](https://github.com/vacp2p/rfc-index/blob/7b443c1aab627894e3f22f5adfbb93f4c4eac4f6/waku/standards/core/12/filter.md#subscriber_ping) to service nodes in use.
+In case those pings fail light node must replace service nodes following advice of [pool of reliable service nodes](./req-res-reliability.md#pool-of-reliable-service-nodes) without waiting for multiple failures.
```
chaitanyaprem, line 131: It would be good to mention that once a node becomes online, it can fetch any missing messages by using store query with timestamps ( from `offlinedetectedTimeStamp`-`few mins` buffer to `onlineTimeStamp`)

chaitanyaprem (approved):
LGTM
Just added 1 comment regarding store queries to get messages when node is offline.

informational/req-res-reliability.md
weboko: I didn't add as this section is about Filter but not against it. Added a note about such option

informational/req-res-reliability.md
```diff
@@ -0,0 +1,139 @@
...
+Information that is important to be up to date:
+- [ENR](../standards/core/enr.md) information;
+- [Libp2p multiaddresses](https://github.com/libp2p/specs/blob/master/addressing/README.md);
+
+### LightPush
+
+#### Sending with redundancy
+To improve chances of delivery of messages light node can attempt sending same message via LightPush to 2 or more service nodes at the same time.
+While doing so it is important to note that bandwidth consumption increases proportionally to amount of additional service nodes used.
+Our advice to use 2 service nodes at a time.
```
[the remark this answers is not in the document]
weboko: this one is actually questionable, let's iterate about it in following update to the RFC?

informational/req-res-reliability.md
```diff
@@ -0,0 +1,139 @@
...
+light node should drop connection to it and a new service node should be connected and added to the pool instead.
+
+We advice to replace service node for LightPush right after first failure in case:
+- connection to it is lost or request timed out;
+- it's response contains [error codes](../standards/core/lightpush.md#examples-of-possible-error-codes): `UNSUPPORTED_PUBSUB_TOPIC`, `INTERNAL_SERVER_ERROR` or `NO_PEERS_TO_RELAY`;
+- request failed but without error message returned;
+
+For Filter we'd recommend replacing service node:
+- [request for subscription](https://github.com/vacp2p/rfc-index/blob/7b443c1aab627894e3f22f5adfbb93f4c4eac4f6/waku/standards/core/12/filter.md#subscribe) so it cannot be initiated;
+- [ping](https://github.com/vacp2p/rfc-index/blob/7b443c1aab627894e3f22f5adfbb93f4c4eac4f6/waku/standards/core/12/filter.md#subscriber_ping) failed 2 times in a row;
```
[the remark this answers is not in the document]
weboko: let's address it in following updates to RFC

I haven't done it because we are actually not sure what is a good number 

currently for js-waku it is a [minute](https://github.com/waku-org/js-waku/blob/c0c2907131374a33de441676de1d38a09993aaab/packages/sdk/src/protocols/filter.ts#L48) but technically we can extend it to two minutes or more due to nwaku dropping subscription after 5-10 minutes

if we do so then some other things come into play - what if network is dropped (mitigated by specifying different behavior for it), nwaku changes behavior etc

so, I think let's see how it works with current number and what we see happening to users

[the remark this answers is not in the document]
weboko: let's revisit in a month or two

informational/req-res-reliability.md
weboko: I think it partially covered with:
https://github.com/waku-org/specs/blob/06692831a68777686c095324c08b33df3fcf7659/informational/req-res-reliability.md?plain=1#L120

but there is still room for improvement if following is true: can a subscription on a service node degrade over time / be dropped with regular pings? if so, is it fixable by just re-creating subscription by init query. 

so, like service node is fine, just code responsible for Filter needs a bit of turn off and on

do you think it is viable, @jm-clius ?

informational/req-res-reliability.md
```diff
@@ -0,0 +1,139 @@
...
+### Filter
+
+#### Regular pings
+To ensure that subscription is maintained by a service node and not closed - light node should do recurring [pings](https://github.com/vacp2p/rfc-index/blob/7b443c1aab627894e3f22f5adfbb93f4c4eac4f6/waku/standards/core/12/filter.md#subscriber_ping).
+Our advice for light node to send ping requests once per minute.
+In case light node does not receive OK response or it times out 2 times - such service node should be replaced as part of maintenance of [pool of reliable service nodes](./req-res-reliability.md#pool-of-reliable-service-nodes).
+Right after such replace light node must create new subscription to newly connected service node as described in [Filter specification](https://github.com/vacp2p/rfc-index/blob/7b443c1aab627894e3f22f5adfbb93f4c4eac4f6/waku/standards/core/12/filter.md).
+
+#### Redundant subscriptions for message loss mitigation
+To mitigate possibility of messages not being delivered by a service node - we advice to consider using multiple Filter subscriptions.
```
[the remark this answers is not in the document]
weboko: dude, it's killing me

informational/req-res-reliability.md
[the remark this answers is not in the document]
jm-clius: Yes, the links on the GH specs are to the github repos and I'd suggest using the persistent links. Afaik, the website renderers replace the links to relative web resources. cc @jimstir to double check.

PR #20, opened 2024-05-29
feat: lightpush new protobuf proposal

NagyZoltanPeter:
Based on https://github.com/waku-org/pm/issues/93 this new version of lightpush protobuf protocol definition is proposed.

- More specific fail case definition in response
   - error status description is now optional rather than hold the factual error cause.
   - error codes are more-or-less matching HTTP error code means
- Response contains the number of successful relay peers.
- Placeholder for request type - for future enhancements

As this is a breaking change compared to the original definition, we may consider to version the protocol allow clients to get adapted to new protocol format.

Open discussion topic: Can't we make pubsub_topic optional for lightpush protocol, indicating to handle message as auto-sharded routing? Currently we require from clients to know about how to compute proper shard from content-topic and make light push request properly with that computed shard topic. Although this can be simple handled on the service side. WDYT?

### DIFF
--- a/standards/core/lightpush.md
+++ b/standards/core/lightpush.md
@@ -1,5 +1,5 @@
 ---
-title: WAKU2-LIGHTPUSH
+title: WAKU-LIGHTPUSH
 name: Waku Light Push
 editor: Zoltan Nagy <zoltan@status.im> 
 contributors: 
--- a/standards/core/lightpush.md
+++ b/standards/core/lightpush.md
@@ -1,7 +1,6 @@
 ---
 title: WAKU2-LIGHTPUSH
-name: Waku v2 Light Push
-status: draft
+name: Waku Light Push
 editor: Zoltan Nagy <zoltan@status.im> 
 contributors: 
   - Hanno Cornelius <hanno@status.im>
@@ -21,7 +20,7 @@ A common use case is to request that service node to publish the message to a `1
 Additionally, there is sometimes a need for confirmation that a message has been received "by the network"
 (here, at least one node).
 
-`WAKU2-LIGHTPUSH` is a request/response protocol for this.
+`WAKU-LIGHTPUSH` is a request/response protocol for this.
 
 ## Payloads
 
@@ -29,9 +28,9 @@ Additionally, there is sometimes a need for confirmation that a message has been
 syntax = "proto3";
 
 message LightPushRequest {
-        string request_id = 1;
-    uin32 request_type = 10;  // reserved for future request type selection, currently it is always RELAY (0)
-    string topic = 20;
+    string request_id = 1;
+    uin32 request_type = 10;  // 10 Reserved for future `request_type`. Currently RELAY is only available service.
+    optional string pubsub_topic = 20;
     WakuMessage message = 21;
 }
 
@@ -45,18 +44,18 @@ message LightPushResponse {
 
 ### Message Relaying
 
-Nodes that respond to `LightPushRequest` MUST either relay the encapsulated message via [11/WAKU2-RELAY](https://rfc.vac.dev/waku/standards/core/11/relay) protocol on the specified `pubsub_topic` or `shard` ([WAKU2-RELAY-SHARDING](https://github.com/waku-org/specs/blob/master/standards/core/relay-sharding.md)) depending on the network configuration.
+Nodes that respond to `LightPushRequest` MUST either relay the encapsulated message via [11/WAKU2-RELAY](https://rfc.vac.dev/waku/standards/core/11/relay) protocol on the specified `pubsub_topic`. Depending on the network configuration, user may not need to provide `pubsub_topic` ([WAKU2-RELAY-SHARDING](https://github.com/waku-org/specs/blob/master/standards/core/relay-sharding.md)).
 If they are unable to do so for some reason, they SHOULD return an error code in `LightPushResponse`.
 
 ### Examples of possible error codes
 
 | Result | Code | Note |
 |--------|------|------|
-| SUCCESS  | 0 | Successfull push, response's relay_peer_count holds the number of peers the message is pushed.    |
-| BAD_REQUEST | 400   |     |
-| NO_PEERS_TO_RELAY | 404 | Service node has no relay peers  |
+| SUCCESS  | 200 | Successfull push, response's relay_peer_count holds the number of peers the message is pushed.    |
+| BAD_REQUEST | 400   | Wrong request payload.    |
+| NO_PEERS_TO_RELAY | 404 | Service node has no relay peers.   |
 | PAYLOAD_TOO_LARGE | 413 | Message exceeds certain size limit, it can depend on network configuration, see status_desc for details.  |
-| UNSUPPORTED_TOPIC | 415 | Requested push on pubsub_topic or shard is not possible as the service node is not subscibed to. |
+| UNSUPPORTED_TOPIC | 421 | Requested push on pubsub_topic is not possible as the service node does not support it. |
 | TOO_MANY_REQUESTS | 429 | DOS protection prevented this request as the current request exceeds the configured request rate. |
 | INTERNAL_SERVER_ERROR  | 500 | status_desc holds explanation of the error.  |
 | SERVICE_UNAVAILABLE | 503 | Lightpush service not available  |
@@ -66,7 +65,7 @@ If they are unable to do so for some reason, they SHOULD return an error code in
 ## Security Considerations
 
 Since this can introduce an amplification factor, it is RECOMMENDED for the node relaying to the rest of the network to take extra precautions.
-Therefore Waku2 applies or will apply:
+Therefore Waku applies or will apply:
 - DOS protection through request rate limitation on the service itself.
 - message rate limiting via [17/WAKU2-RLN-RELAY](https://rfc.vac.dev/waku/standards/core/17/rln-relay), applied via network membership subscription.
 
--- a/standards/core/lightpush.md
+++ b/standards/core/lightpush.md
@@ -53,12 +53,11 @@ If they are unable to do so for some reason, they SHOULD return an error code in
 |--------|------|------|
 | SUCCESS  | 200 | Successfull push, response's relay_peer_count holds the number of peers the message is pushed.    |
 | BAD_REQUEST | 400   | Wrong request payload.    |
-| NO_PEERS_TO_RELAY | 404 | Service node has no relay peers.   |
 | PAYLOAD_TOO_LARGE | 413 | Message exceeds certain size limit, it can depend on network configuration, see status_desc for details.  |
 | UNSUPPORTED_TOPIC | 421 | Requested push on pubsub_topic is not possible as the service node does not support it. |
 | TOO_MANY_REQUESTS | 429 | DOS protection prevented this request as the current request exceeds the configured request rate. |
 | INTERNAL_SERVER_ERROR  | 500 | status_desc holds explanation of the error.  |
-| SERVICE_UNAVAILABLE | 503 | Lightpush service not available  |
+| NO_PEERS_TO_RELAY | 503 | Lightpush service not available as node has no relay peers. |
 
 > The list of error codes are not complete and can be extended in the future.
 
--- a/standards/core/lightpush.md
+++ b/standards/core/lightpush.md
@@ -29,7 +29,7 @@ syntax = "proto3";
 
 message LightPushRequest {
     string request_id = 1;
-    uin32 request_type = 10;  // 10 Reserved for future `request_type`. Currently, RELAY is the only available service.
+    // 10 Reserved for future `request_type`. Currently, RELAY is the only available service.
     optional string pubsub_topic = 20;
     WakuMessage message = 21;
 }
@@ -44,10 +44,13 @@ message LightPushResponse {
 
 ### Message Relaying
 
-Nodes that respond to `LightPushRequest` MUST either relay the encapsulated message via [11/WAKU2-RELAY](https://rfc.vac.dev/waku/standards/core/11/relay) protocol on the specified `pubsub_topic`. Depending on the network configuration, the user may not need to provide `pubsub_topic` ([WAKU2-RELAY-SHARDING](https://github.com/waku-org/specs/blob/master/standards/core/relay-sharding.md)).
+Nodes that respond to `LightPushRequest` SHOULD either relay the encapsulated message via [11/WAKU2-RELAY](https://rfc.vac.dev/waku/standards/core/11/relay) protocol on the specified `pubsub_topic`
+or perform another requested service.
+Services beyond [11/WAKU2-RELAY](https://rfc.vac.dev/waku/standards/core/11/relay) are as yet underdefined.
+Depending on the network configuration, the lightpush client may not need to provide `pubsub_topic` ([WAKU2-RELAY-SHARDING](https://github.com/waku-org/specs/blob/master/standards/core/relay-sharding.md)).
 If the node is unable to do so for some reason, they SHOULD return an error code in `LightPushResponse`.
 Once the relay is successful, the `relay_peer_count` will indicate the number of peers that the node has managed to relay the message to. It's important to note that this number may vary depending on the node subscriptions and support for the requested pubsub_topic. The client can use this information to either consider the relay as successful or take further action, such as switching to a lightpush service peer with better connectivity.
-The field `relay_peer_count` may not present or has the value zero in case of error or in other future use cases, where no relay is involved.
+The field `relay_peer_count` may not be present or has the value zero in case of error or in other future use cases, where no relay is involved.
 
 ### Examples of possible error codes
 
--- a/standards/core/lightpush.md
+++ b/standards/core/lightpush.md
@@ -15,7 +15,7 @@ previous version: `/vac/waku/lightpush/2.0.0-beta1` [19/WAKU2-LIGHTPUSH](https:/
 
 ## Motivation and Goals
 
-Light nodes with short connection windows and limited bandwidth wish to push messages to other nodes in the Waku network to request message services.
+Light nodes with short connection windows and limited bandwidth wish to push messages to other nodes in the Waku network to request message services.<br>
 A common use case is to request that the service node publish the message to an `11/WAKU2-RELAY` pubsub-topic.
 Additionally, there is sometimes a need for confirmation that a message has been received "by the network"
 (here, at least one node).
@@ -44,13 +44,17 @@ message LightPushResponse {
 
 ### Message Relaying
 
-Nodes that respond to `LightPushRequest` SHOULD either relay the encapsulated message via [11/WAKU2-RELAY](https://rfc.vac.dev/waku/standards/core/11/relay) protocol on the specified `pubsub_topic`
-or perform another requested service.
-Services beyond [11/WAKU2-RELAY](https://rfc.vac.dev/waku/standards/core/11/relay) are as yet underdefined.
-Depending on the network configuration, the lightpush client may not need to provide `pubsub_topic` ([WAKU2-RELAY-SHARDING](https://github.com/waku-org/specs/blob/master/standards/core/relay-sharding.md)).
-If the node is unable to do so for some reason, they SHOULD return an error code in `LightPushResponse`.
-Once the relay is successful, the `relay_peer_count` will indicate the number of peers that the node has managed to relay the message to. It's important to note that this number may vary depending on the node subscriptions and support for the requested pubsub_topic. The client can use this information to either consider the relay as successful or take further action, such as switching to a lightpush service peer with better connectivity.
-The field `relay_peer_count` may not be present or has the value zero in case of error or in other future use cases, where no relay is involved.
+Nodes that respond to `LightPushRequest` SHOULD
+- either relay the encapsulated message via [11/WAKU2-RELAY](https://rfc.vac.dev/waku/standards/core/11/relay) protocol on the specified `pubsub_topic`<br>
+- or perform another requested service.
+  `Services beyond [11/WAKU2-RELAY](https://rfc.vac.dev/waku/standards/core/11/relay) are as yet underdefined.`
+
+Depending on the network configuration, the lightpush client may not need to provide `pubsub_topic` ([WAKU2-RELAY-SHARDING](https://github.com/waku-org/specs/blob/master/standards/core/relay-sharding.md)).<br>
+
+If the node is unable to perform the request for some reason, it SHOULD return an error code in `LightPushResponse`.<br>
+
+Once the relay is successful, the `relay_peer_count` will indicate the number of peers that the node has managed to relay the message to. It's important to note that this number may vary depending on the node subscriptions and support for the requested pubsub_topic. The client can use this information to either consider the relay as successful or take further action, such as switching to a lightpush service peer with better connectivity.<br>
+>The field `relay_peer_count` may not be present or has the value zero in case of error or in other future use cases, where no relay is involved.
 
 ### Examples of possible error codes
 
--- a/standards/core/lightpush.md
+++ b/standards/core/lightpush.md
@@ -45,7 +45,7 @@ message LightPushResponse {
 ### Message Relaying
 
 Nodes that respond to `LightPushRequest` MUST either relay the encapsulated message via [11/WAKU2-RELAY](https://rfc.vac.dev/waku/standards/core/11/relay) protocol on the specified `pubsub_topic`. Depending on the network configuration, the user may not need to provide `pubsub_topic` ([WAKU2-RELAY-SHARDING](https://github.com/waku-org/specs/blob/master/standards/core/relay-sharding.md)).
-If they are unable to do so for some reason, they SHOULD return an error code in `LightPushResponse`.
+If the node is unable to do so for some reason, they SHOULD return an error code in `LightPushResponse`.
 Once the relay is successful, the `relay_peer_count` will indicate the number of peers that the node has managed to relay the message to. It's important to note that this number may vary depending on the node subscriptions and support for the requested pubsub_topic. The client can use this information to either consider the relay as successful or take further action, such as switching to a lightpush service peer with better connectivity.
 
 ### Examples of possible error codes
@@ -55,7 +55,7 @@ Once the relay is successful, the `relay_peer_count` will indicate the number of
 | SUCCESS  | 200 | Successfull push, response's relay_peer_count holds the number of peers the message is pushed.    |
 | BAD_REQUEST | 400   | Wrong request payload.    |
 | PAYLOAD_TOO_LARGE | 413 | Message exceeds certain size limit, it can depend on network configuration, see status_desc for details.  |
-| UNSUPPORTED_TOPIC | 421 | Requested push on pubsub_topic is not possible as the service node does not support it. |
+| UNSUPPORTED_PUBSUB_TOPIC | 421 | Requested push on pubsub_topic is not possible as the service node does not support it. |
 | TOO_MANY_REQUESTS | 429 | DOS protection prevented this request as the current request exceeds the configured request rate. |
 | INTERNAL_SERVER_ERROR  | 500 | status_desc holds explanation of the error.  |
 | NO_PEERS_TO_RELAY | 503 | Lightpush service is not available as the node has no relay peers. |
--- a/standards/core/lightpush.md
+++ b/standards/core/lightpush.md
@@ -1,6 +1,5 @@
 ---
-slug: 19
-title: 19/WAKU2-LIGHTPUSH
+title: WAKU2-LIGHTPUSH
 name: Waku v2 Light Push
 status: draft
 editor: Zoltan Nagy <zoltan@status.im> 
@@ -13,15 +12,16 @@ contributors:
 previous version: `/vac/waku/lightpush/2.0.0-beta1` [19/WAKU2-LIGHTPUSH](https://rfc.vac.dev/waku/standards/core/19/lightpush)
 
 ---
-**Protocol identifier**: `/vac/waku/lightpush/2.0.0-beta2`
+**Protocol identifier**: `/vac/waku/lightpush/3.0.0`
 
 ## Motivation and Goals
 
-Light nodes with short connection windows and limited bandwidth wish to publish messages into the Waku network.
+Light nodes with short connection windows and limited bandwidth wish to push messages to other nodes in the Waku network to request message services.
+A common use case is to request that service node to publish the message to a `11/WAKU2-RELAY` pubsub topic.
 Additionally, there is sometimes a need for confirmation that a message has been received "by the network"
 (here, at least one node).
 
-`19/WAKU2-LIGHTPUSH` is a request/response protocol for this.
+`WAKU2-LIGHTPUSH` is a request/response protocol for this.
 
 ## Payloads
 
@@ -29,30 +29,15 @@ Additionally, there is sometimes a need for confirmation that a message has been
 syntax = "proto3";
 
 message LightPushRequest {
-    enum RequestType { // Future extension point
-        RELAY = 0;
-    }
-    
-    string request_id = 1;
-    RequestType kind = 10;  // default is RELAY, placeholder for future extensions
-    string pubsub_topic = 20;
+        string request_id = 1;
+    uin32 request_type = 10;  // reserved for future request type selection, currently it is always RELAY (0)
+    string topic = 20;
     WakuMessage message = 21;
 }
 
 message LightPushResponse {
-    enum Status {
-        SUCCESS = 0;
-        BAD_REQUEST = 400;
-        NO_PEERS_TO_RELAY = 404;
-        PAYLOAD_TOO_LARGE = 413;
-        UNSUPPORTED_TOPIC = 415;
-        TOO_MANY_REQUESTS = 429;
-        INTERNAL_SERVER_ERROR = 500;
-        SERVICE_UNAVAILABLE = 503;    
-    }
-
     string request_id = 1;
-    Status status_code = 10;
+    uint32 status_code = 10; // non zero in case of failure, see appendix
     optional string status_desc = 11;
     uint32 relay_peer_count = 12;
 }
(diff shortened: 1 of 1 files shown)


standards/core/lightpush.md
```diff
@@ -0,0 +1,88 @@
...
+contributors: 
+  - Hanno Cornelius <hanno@status.im>
+  - Daniel Kaiser <danielkaiser@status.im>
+  - Oskar Thorén <oskarth@titanproxy.com>
+---
+
+previous version: `/vac/waku/lightpush/2.0.0-beta1` [19/WAKU2-LIGHTPUSH](https://rfc.vac.dev/waku/standards/core/19/lightpush)
+
+---
+**Protocol identifier**: `/vac/waku/lightpush/2.0.0-beta2`
```
SionoiS, line 16: pet peeve of mine but I hate "beta" labels. Just increase the number! :stuck_out_tongue:

standards/core/lightpush.md
```diff
@@ -0,0 +1,88 @@
...
+syntax = "proto3";
+
+message LightPushRequest {
+    enum RequestType { // Future extension point
+        RELAY = 0;
+    }
+    
+    string request_id = 1;
+    RequestType kind = 10;  // default is RELAY, placeholder for future extensions
+    string pubsub_topic = 20;
```
SionoiS, line 38: pubsub topic would be the most agnostic but maybe we could use shard instead?

standards/core/lightpush.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Payloads
+
+```protobuf
+syntax = "proto3";
+
+message LightPushRequest {
+    enum RequestType { // Future extension point
+        RELAY = 0;
+    }
```
SionoiS, line 34: In my mind we push a message to a node not to a protocol.

The only other protocol I can think of would be store.

:thinking: What would it mean to push to store instead of relay?

standards/core/lightpush.md
```diff
@@ -0,0 +1,88 @@
...
+    enum Status {
+        SUCCESS = 0;
+        BAD_REQUEST = 400;
+        NO_PEERS_TO_RELAY = 404;
+        PAYLOAD_TOO_LARGE = 413;
+        UNSUPPORTED_TOPIC = 415;
+        TOO_MANY_REQUESTS = 429;
+        INTERNAL_SERVER_ERROR = 500;
+        SERVICE_UNAVAILABLE = 503;    
+    }
```
SionoiS, line 52: What if instead of defining the codes we keep this impl. detail? We can still specify that error codes should follow http standards.

standards/core/lightpush.md
```diff
@@ -0,0 +1,88 @@
...
+
+    string request_id = 1;
+    Status status_code = 10;
+    optional string status_desc = 11;
+    uint32 relay_peer_count = 12;
```
SionoiS, line 57: What if instead of sending back this number. We require the sender to specify the minimum # of peers sent to that they feel is appropriate?

standards/core/lightpush.md
```diff
@@ -0,0 +1,88 @@
...
+}
+```
+
+### Message Relaying
+
+Nodes that respond to `LightPushRequest` MUST either
+relay the encapsulated message via [11/WAKU2-RELAY](https://rfc.vac.dev/waku/standards/core/11/relay) protocol on the specified `pubsub_topic`,
+or forward the `LightPushRequest` via 19/LIGHTPUSH on a [WAKU2-DANDELION](https://github.com/waku-org/specs/blob/waku-RFC/standards/application/dandelion.md) stem.
```
SionoiS, line 65: Is this still relevant?

standards/core/lightpush.md
```diff
@@ -0,0 +1,88 @@
...
+## Security Considerations
+
+Since this can introduce an amplification factor, it is RECOMMENDED for the node relaying to the rest of the network to take extra precautions.
+This can be done by rate limiting via [17/WAKU2-RLN-RELAY](https://rfc.vac.dev/waku/standards/core/17/rln-relay).
+
+Note that the above is currently not fully implemented.
+
+## Future work
+
+- Add support attaching RLN proof for the message requested to be relayed.
```
SionoiS, line 77: I guess to be 100% future proof we should add this functionality.

standards/core/lightpush.md
```diff
@@ -0,0 +1,88 @@
...
+relay the encapsulated message via [11/WAKU2-RELAY](https://rfc.vac.dev/waku/standards/core/11/relay) protocol on the specified `pubsub_topic`,
+or forward the `LightPushRequest` via 19/LIGHTPUSH on a [WAKU2-DANDELION](https://github.com/waku-org/specs/blob/waku-RFC/standards/application/dandelion.md) stem.
+If they are unable to do so for some reason, they SHOULD return an error code in `LightPushResponse`.
+
+## Security Considerations
+
+Since this can introduce an amplification factor, it is RECOMMENDED for the node relaying to the rest of the network to take extra precautions.
+This can be done by rate limiting via [17/WAKU2-RLN-RELAY](https://rfc.vac.dev/waku/standards/core/17/rln-relay).
+
+Note that the above is currently not fully implemented.
```
SionoiS, line 73: It is now :smile:

standards/core/lightpush.md
NagyZoltanPeter: @jm-clius  can explain I think, but mainly yes, push message to store but not relay or both and additionally the RFC refers to 

```forward the `LightPushRequest` via 19/LIGHTPUSH on a [WAKU2-DANDELION](https://github.com/waku-org/specs/blob/waku-RFC/standards/application/dandelion.md) stem```

File: standards/core/lightpush.md (hunk 4 above)
Comment (NagyZoltanPeter) [reply standards/core/lightpush.md:52]: Hm, interesting idea not to have them in the communication interface description what outcome you might prepare for. For me it is more should be here than not.

File: standards/core/lightpush.md (hunk 3 above)
Comment (gabrielmer) [reply standards/core/lightpush.md:34]: Wouldn't pushing to Relay with the `ephemeral` field on be the same as pushing to both Relay and Store?

Comment (NagyZoltanPeter) [reply standards/core/lightpush.md:34]: How I understood this, to request just a store of the message but not to relay. Suppose we have a light client where using a lightpush service to directly store a message. @jm-clius is this what you meant?

File: standards/core/lightpush.md [hunk 9]
```diff
@@ -0,0 +1,88 @@
...
+    WakuMessage message = 21;
+}
+
+message LightPushResponse {
+    enum Status {
+        SUCCESS = 0;
+        BAD_REQUEST = 400;
+        NO_PEERS_TO_RELAY = 404;
+        PAYLOAD_TOO_LARGE = 413;
+        UNSUPPORTED_TOPIC = 415;
```
alrevuelta, line 48: would use contenttopic or pubsubtopic to avoid confusion.

standards/core/lightpush.md
```diff
@@ -0,0 +1,88 @@
...
+        TOO_MANY_REQUESTS = 429;
```
alrevuelta, line 49: Guess this maps to rln? Unsure how deep we could go but rln could lead to multiple errors. Some ideas:
* too many requests for epoch. assuming rlnv2 this means that the user has sent more messages than allowed within a given epoch.
* wrong epoch (either to early or to late)
* invalid root. the attached root of the proof does not match the window of acceptable roots.
* invalid proof. perhaps more generic to consider invalid proofs.

standards/core/lightpush.md
alrevuelta: using shard can work, but guess we would need the cluster-id as well. since the pubsub topic encodes both.?

alrevuelta (commented):
thanks, left some comments.

standards/core/lightpush.md
```diff
@@ -0,0 +1,88 @@
+---
+slug: 19
+title: 19/WAKU2-LIGHTPUSH
```
jm-clius, line 3: No need for a number or slug. It can simply be titled WAKU2-LIGHTPUSH. Once implemented and tested, it could graduate to a Vac RFC and will get the number 19 then to indicate it replaces the current published spec.

standards/core/lightpush.md
jm-clius: Agree. We can just call it 3.0.0

standards/core/lightpush.md
```diff
@@ -0,0 +1,88 @@
...
+---
+
+previous version: `/vac/waku/lightpush/2.0.0-beta1` [19/WAKU2-LIGHTPUSH](https://rfc.vac.dev/waku/standards/core/19/lightpush)
+
+---
+**Protocol identifier**: `/vac/waku/lightpush/2.0.0-beta2`
+
+## Motivation and Goals
+
+Light nodes with short connection windows and limited bandwidth wish to publish messages into the Waku network.
```
jm-clius, line 20: Nitpick, but perhaps we can change the phrasing here in such a way that we do not commit to only one use case for lightpush, e.g.

```suggestion
Light nodes with short connection windows and limited bandwidth wish to push messages to other nodes in the Waku network to request message services.
A common use case is to request that service node to publish the message to a `11/WAKU2-RELAY` pubsub topic.
```

standards/core/lightpush.md
jm-clius: > In my mind we push a message to a node not to a protocol.

Indeed, I think the point here is to allow for future differentiation of "message services" that you're requesting from the service node. In that sense it's not a list of protocols in the request type, but a list of possible services of which "relay publishing" may be one. Other possible services that we envision:
- store the message (without publishing to relay)
- dandelion hop (for anonymisation)
- add rln proof and publish (arguably a different service than just publishing)

However, @NagyZoltanPeter I think for this version of the protocol we can perhaps leave the enum undefined and just add a comment that tag/field number `10` is reserved for this use case? It would make more sense to add once we have more services.

standards/core/lightpush.md
jm-clius: Indeed, for new versions of protocols (and after a number of discussions), we've opted to leave `status_code` and `status_desc` generic `uint32` and `string` fields in order to define the possible error codes in the implementation. The reason being that we extend this list quite often and we'll soon get in a situation where the implementation and spec diverge. Of course, once we believe we have some stability in the services, we may publish a stand-beside spec with the list of supported error codes. See https://github.com/vacp2p/rfc-index/blob/main/waku/standards/core/12/filter.md#protobuf for how we're doing it in the new filter (and new store) protocols.

standards/core/lightpush.md
jm-clius: I would remove dandelion reference here (afaik it's not implemented) and just rephrase that by default the service provided is RELAY (without committing to this being the only way we'll process messages in future).

jm-clius (changes requested):
Thanks! A couple of comments below, but direction lgtm

standards/core/lightpush.md
NagyZoltanPeter: Right, changed.

standards/core/lightpush.md
NagyZoltanPeter: Ok, removed

standards/core/lightpush.md
NagyZoltanPeter: Not exactly TOO_MANY_REQUESTS relates to rate limiting the service itself.
Bu I do agree, we can add more error codes that related to RLN. 
Also as agreed to move out the exact list of possible outcomes from the protocol definition it will be more easy to add any new if needed.

standards/core/lightpush.md
NagyZoltanPeter: Er, what is the suggestion exactly? Maybe we can use different wording.
It refers to such requests having pubsub topic our service node is not subscribed to.
Of course it depends on the use case and configuration, what type of sharding is in use.

standards/core/lightpush.md
NagyZoltanPeter: it is currently on going by Akhil.

standards/core/lightpush.md
NagyZoltanPeter: I rephrased the section.

standards/core/lightpush.md
NagyZoltanPeter: Done. Added into a separate section as examples.

standards/core/lightpush.md
NagyZoltanPeter: I agree but IDK which wording would be the most straightforward to use. Do we have any in use?
Maybe `topic` as simple as is.

NagyZoltanPeter: WDYT? @alrevuelta, @jm-clius, @SionoiS, @Ivansete-status

standards/core/lightpush.md
```diff
@@ -0,0 +1,88 @@
+---
+title: WAKU2-LIGHTPUSH
+name: Waku v2 Light Push
```
Ivansete-status, line 3: I think we can start avoiding using the "2" term and simply refer to Waku
```suggestion
title: WAKU-LIGHTPUSH
name: Waku Light Push
```

standards/core/lightpush.md
Ivansete-status: IMO, the `lightPush` should not care about the `relay` connected peers. We can simply return an "LOW_NUMBER_OF_PEERS" error which will make the lightpush client to just try with another relay peer. The current number of relay peers can be returned back within the `status_desc` field

standards/core/lightpush.md
```diff
@@ -0,0 +1,88 @@
...
+**Protocol identifier**: `/vac/waku/lightpush/3.0.0`
+
+## Motivation and Goals
+
+Light nodes with short connection windows and limited bandwidth wish to push messages to other nodes in the Waku network to request message services.
+A common use case is to request that service node to publish the message to a `11/WAKU2-RELAY` pubsub topic.
+Additionally, there is sometimes a need for confirmation that a message has been received "by the network"
+(here, at least one node).
+
+`WAKU2-LIGHTPUSH` is a request/response protocol for this.
```
Ivansete-status, line 24:
```suggestion
`WAKU-LIGHTPUSH` is a request/response protocol for this.
```

standards/core/lightpush.md
```diff
@@ -0,0 +1,88 @@
...
+Nodes that respond to `LightPushRequest` MUST either relay the encapsulated message via [11/WAKU2-RELAY](https://rfc.vac.dev/waku/standards/core/11/relay) protocol on the specified `pubsub_topic` or `shard` ([WAKU2-RELAY-SHARDING](https://github.com/waku-org/specs/blob/master/standards/core/relay-sharding.md)) depending on the network configuration.
+If they are unable to do so for some reason, they SHOULD return an error code in `LightPushResponse`.
+
+### Examples of possible error codes
+
+| Result | Code | Note |
+|--------|------|------|
+| SUCCESS  | 0 | Successfull push, response's relay_peer_count holds the number of peers the message is pushed.    |
+| BAD_REQUEST | 400   |     |
+| NO_PEERS_TO_RELAY | 404 | Service node has no relay peers  |
```
Ivansete-status, line 57: A possible suggestion because the `NO_PEERS_TO_RELAY` value has a connotation of zero peers.
```suggestion
| NOT_ENOUGH_PEERS_TO_RELAY | 404 | Service node has no relay peers  |
```

standards/core/lightpush.md
```diff
@@ -0,0 +1,88 @@
...
+| TOO_MANY_REQUESTS | 429 | DOS protection prevented this request as the current request exceeds the configured request rate. |
+| INTERNAL_SERVER_ERROR  | 500 | status_desc holds explanation of the error.  |
+| SERVICE_UNAVAILABLE | 503 | Lightpush service not available  |
+
+> The list of error codes are not complete and can be extended in the future.
+
+## Security Considerations
+
+Since this can introduce an amplification factor, it is RECOMMENDED for the node relaying to the rest of the network to take extra precautions.
+Therefore Waku2 applies or will apply:
```
Ivansete-status, line 69:
```suggestion
Therefore Waku applies or will apply:
```

Ivansete-status (commented):
Thanks for it! :partying_face: 
Just added some comments so far that I hope you find useful

NagyZoltanPeter:
https://github.com/waku-org/specs/pull/20#discussion_r1632370661
@Ivansete-status: I think its better idea to put less such logic into the service side. I feel it more client side one. IMHO we might embed it into SDK rather than to the raw protocol. 
Because what does LOW_NUMBER_OF_PEERS implies? Is it a fixed number by configuration which the client side might not know about or depends on other factors like due to bandwidth restriction we might decide to unsubscribe from some of the heavy load shards? 
Counting with DOS protection, following bandwidth management and other features may come (e.g. light push request type: push to store) that will impose other considerations.

Ivansete-status:
> [#20 (comment)](https://github.com/waku-org/specs/pull/20#discussion_r1632370661) @Ivansete-status: I think its better idea to put less such logic into the service side. I feel it more client side one. IMHO we might embed it into SDK rather than to the raw protocol.

hahah yes I've noticed that :) The following comment is related to that and why I consider we should have fewer assumptions in light-node side and more on the full-node side: https://github.com/waku-org/nwaku/issues/1004#issuecomment-2156807537

-------

> Because what does LOW_NUMBER_OF_PEERS implies? Is it a fixed number by configuration which the client side might not know about or depends on other factors like due to bandwidth restriction we might decide to unsubscribe from some of the heavy load shards?

 `<<Is it a fixed number by configuration which the client side might not know about>>` --> Yes exactly. In that case, the light node may decide to try again to the same full-node in the near future or try to use another full-node.

---- 
> Counting with DOS protection, following bandwidth management and other features may come (e.g. light push request type: push to store) that will impose other considerations.

Sorry, but I can't quite follow that statement.

standards/core/lightpush.md
SionoiS: > > In my mind we push a message to a node not to a protocol.
> 
> Indeed, I think the point here is to allow for future differentiation of "message services" that you're requesting from the service node. In that sense it's not a list of protocols in the request type, but a list of possible services of which "relay publishing" may be one. Other possible services that we envision:
> 
> * store the message (without publishing to relay)
> * dandelion hop (for anonymisation)
> * add rln proof and publish (arguably a different service than just publishing)

I re-read and I feel like it those use cases could be new protocols not extension. What's the benefit of doing it this way? The obvious downside is the added complexity of having multiple "type" of services in one protocol.

SionoiS (changes requested):
LGTM except `topic` see my comment

standards/core/lightpush.md
```diff
@@ -0,0 +1,88 @@
...
+`WAKU2-LIGHTPUSH` is a request/response protocol for this.
+
+## Payloads
+
+```protobuf
+syntax = "proto3";
+
+message LightPushRequest {
+        string request_id = 1;
+    uin32 request_type = 10;  // reserved for future request type selection, currently it is always RELAY (0)
```
jm-clius, line 33: Should be `uint32`. I would comment out this field completely for now though. Something like:
```suggestion
    // 10 Reserved for future `request_type`. Currently RELAY is only available service.
```

standards/core/lightpush.md
```diff
@@ -0,0 +1,88 @@
...
+    string topic = 20;
```
jm-clius, line 34:
```suggestion
    optional string pubsub_topic = 20;
```
We should be explicit about this being a `pubsub_topic`. Note that making the field `optional` here doesn't mean that it is optional in the spec - it will simply allow the service node to verify that a `pubsub_topic` has been explicitly set by the client. Otherwise it will by default interpret any message withouth `pubsub_topic` as being directed at the empty string pubsub topic.

standards/core/lightpush.md
jm-clius: I think it makes sense for the service node to return the number of peers and the client to decide what to do with this information.

standards/core/lightpush.md
```diff
@@ -0,0 +1,88 @@
...
+    string request_id = 1;
+    uint32 status_code = 10; // non zero in case of failure, see appendix
+    optional string status_desc = 11;
+    uint32 relay_peer_count = 12;
+}
+```
+
+### Message Relaying
+
+Nodes that respond to `LightPushRequest` MUST either relay the encapsulated message via [11/WAKU2-RELAY](https://rfc.vac.dev/waku/standards/core/11/relay) protocol on the specified `pubsub_topic` or `shard` ([WAKU2-RELAY-SHARDING](https://github.com/waku-org/specs/blob/master/standards/core/relay-sharding.md)) depending on the network configuration.
```
jm-clius, line 48: I would avoid using `shard` on this layer. The use of "either" here doesn't make sense as we don't provide an alternative.

standards/core/lightpush.md
jm-clius: I'd suggest sticking to HTTP codes as much as possible. A `404` usually implies an addressed resource was "not found" - a mistake the client needs to fix. Wouldn't "no peers to relay" match more closely to the "service unavailable" `503`? That is, you requested a service, but it is temporarily unavailable on the server-side due to lack of peers to relay to. We could in theory also use some of the unused HTTP codes to have more specific error cases.
Generally server-side errors should be in the `500` range.
Client-side errors (e.g. addressing a missing resource) should be in the `400` range.

See e.g. https://developer.mozilla.org/en-US/docs/Web/HTTP/Status

standards/core/lightpush.md
```diff
@@ -0,0 +1,88 @@
+---
+title: WAKU2-LIGHTPUSH
+name: Waku v2 Light Push
+status: draft
```
jm-clius, line 4: For a new protocol, remove the `status`
```suggestion
```

jm-clius (approved):
Approving as I will be AFK for a couple of days and don't want to be a bottleneck. A couple of comments below. :)

standards/core/lightpush.md
alrevuelta: `UNSUPPORTED_TOPIC` -> `UNSUPPORTED_PUBSUBTOPIC`

standards/core/lightpush.md
weboko: Agree with @jm-clius and I think we should have: 
- SUCCESS - 200;
- BAD_REQUEST - to me is questionable, do we even need it? 
- UNSUPPORTED_TOPIC - maybe `421`?
- SERVICE_UNAVAILABLE overlaps with NO_PEERS_TO_RELAY - seems to me too;

Also, we need to figure out how to apply `418`! I always dreamt of doing so. 

cc @NagyZoltanPeter

weboko (approved):
looks good and added some comments

standards/core/lightpush.md
NagyZoltanPeter: Done

standards/core/lightpush.md
NagyZoltanPeter: Thanks.

standards/core/lightpush.md
NagyZoltanPeter: Right, thank you.

standards/core/lightpush.md
NagyZoltanPeter: I would close this as is here, returning exact peer count the message had been relayed to, gives the chance and flexibility to the client or SDK to decide its reactions. Other options would make the protocol just even fluid and harder to implement properly for other use cases.

standards/core/lightpush.md
NagyZoltanPeter: I reformatted this section.

standards/core/lightpush.md
NagyZoltanPeter: @weboko: I'm not sure that having non-zero success helps here, but I will do it to get it confirm with STORE's status/error codes.
418? :-D maybe for BAD_REQUEST, that's the closest, haha.

NagyZoltanPeter: Oh, and yes, BAD_REQUEST is anything if we cannot decode the request's payoad. It can be due to missused (wrong) protobuf definition.

NagyZoltanPeter: @weboko: For `SERVICE_UNAVAILABLE overlaps with NO_PEERS_TO_RELAY` question. I'm not pretty sure. There can be different reason why we don't have peers.
I'm thinking of: SERVICE_UNAVAILABLE - after startup we did not get connected to any peers yet, or we are disconnected for some reason. While NO_PEERS_TO_RELAY is rather related to the used pubsub_topic/shard we do not have connected peers that supports that. WDYT? I would keep both in this form.

standards/core/lightpush.md
NagyZoltanPeter: Done

standards/core/lightpush.md
```diff
@@ -0,0 +1,88 @@
...
+
+### Examples of possible error codes
+
+| Result | Code | Note |
+|--------|------|------|
+| SUCCESS  | 0 | Successfull push, response's relay_peer_count holds the number of peers the message is pushed.    |
+| BAD_REQUEST | 400   |     |
+| NO_PEERS_TO_RELAY | 404 | Service node has no relay peers  |
+| PAYLOAD_TOO_LARGE | 413 | Message exceeds certain size limit, it can depend on network configuration, see status_desc for details.  |
+| UNSUPPORTED_TOPIC | 415 | Requested push on pubsub_topic or shard is not possible as the service node is not subscibed to. |
```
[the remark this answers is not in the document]
NagyZoltanPeter: If that can be known than it is not a leak of info, isn't it? But can be valuable information for lightpush-client / SDK to decide to whether use or not use a service peer.
I'm also thinking of future planned dynamic bandwidth management feature where a node may decide to unsubscribe from some shards in order to lower resource needs (based on config).

NagyZoltanPeter:
@weboko , @SionoiS, @Ivansete-status : Please take a last look if acceptable. I'm heading to merge this PR. It's time to make some coding soon. :-)

NagyZoltanPeter:
I changed my mind as was thinking of comments from @jm-clius and @weboko... https://github.com/waku-org/specs/pull/20#discussion_r1638422709

I removed SERVICE_UNAVAILABLE and replaced with NO_PEERS_TO_RELAY as in this particular application for lightpush it means exactly that.

standards/core/lightpush.md
```diff
@@ -0,0 +1,86 @@
+---
+title: WAKU2-LIGHTPUSH
```
Ivansete-status, line 2: Small comment: I think we should start only using the "waku" term. I see is actually well used in line 23.
```suggestion
title: WAKU-LIGHTPUSH
```

standards/core/lightpush.md
Ivansete-status: @NagyZoltanPeter - I think it would be interesting to make it clear what the `relay_peer_count` means: whether it represents the actual number of connected nodes or the number of nodes that form the mesh. I'm happy with the decision but we might need to add a small comment somewhere to clarify that (even a comment next to this definition would be great :) )

( cc @jm-clius )

Ivansete-status (approved):
LGTM! Thanks so much for it and the patience :raised_hands: !
I'm just leaving a couple of minor comments.
From my PoV we need to make it a bit clearer the meaning of `relay_peer_count`.

NagyZoltanPeter:
@Ivansete-status yes you are right, I will lately extent explanation of relay peer count as it may not matching mesh node count.

standards/core/lightpush.md
NagyZoltanPeter: Dandelion ref is removed.

standards/core/lightpush.md
NagyZoltanPeter: Renamed.

standards/core/lightpush.md
NagyZoltanPeter: For a relay network it is still topic or pubsub_topic which routes the message, shard has meaning only under waku and not explained here well. I put a link there to direct interest.

NagyZoltanPeter:
@SionoiS: Please take a look and approve if it is now in acceptable form. Thank you.

standards/core/lightpush.md
```diff
@@ -0,0 +1,87 @@
...
+    uin32 request_type = 10;  // 10 Reserved for future `request_type`. Currently, RELAY is the only available service.
+    optional string pubsub_topic = 20;
+    WakuMessage message = 21;
+}
+
+message LightPushResponse {
+    string request_id = 1;
+    uint32 status_code = 10; // non zero in case of failure, see appendix
+    optional string status_desc = 11;
+    uint32 relay_peer_count = 12; // number of peers, message is successfully relayed to
```
chaitanyaprem, line 41: Shouldn't this be optional? As this is specific to RELAY. In future if we add some other request_type, this field may not be required in the response.

standards/core/lightpush.md
```diff
@@ -0,0 +1,88 @@
...
+`WAKU-LIGHTPUSH` is a request/response protocol for this.
+
+## Payloads
+
+```protobuf
+syntax = "proto3";
+
+message LightPushRequest {
+    string request_id = 1;
+    uin32 request_type = 10;  // 10 Reserved for future `request_type`. Currently, RELAY is the only available service.
```
jm-clius, line 32:
```suggestion
    // 10 Reserved for future `request_type`. Currently, RELAY is the only available service.
```

I'd still leave this field out completely. In other words, add the comment but not the field.

standards/core/lightpush.md
```diff
@@ -0,0 +1,88 @@
...
+    string request_id = 1;
+    uint32 status_code = 10; // non zero in case of failure, see appendix
+    optional string status_desc = 11;
+    optional uint32 relay_peer_count = 12; // number of peers, message is successfully relayed to 
+}
+```
+
+### Message Relaying
+
+Nodes that respond to `LightPushRequest` MUST either relay the encapsulated message via [11/WAKU2-RELAY](https://rfc.vac.dev/waku/standards/core/11/relay) protocol on the specified `pubsub_topic`. Depending on the network configuration, the user may not need to provide `pubsub_topic` ([WAKU2-RELAY-SHARDING](https://github.com/waku-org/specs/blob/master/standards/core/relay-sharding.md)).
```
jm-clius, line 47:
```suggestion
Nodes that respond to `LightPushRequest` SHOULD either relay the encapsulated message via [11/WAKU2-RELAY](https://rfc.vac.dev/waku/standards/core/11/relay) protocol on the specified `pubsub_topic`
or perform another requested service.
Services beyond [11/WAKU2-RELAY](https://rfc.vac.dev/waku/standards/core/11/relay) are as yet underdefined.
Depending on the network configuration, the lightpush client may not need to provide `pubsub_topic` ([WAKU2-RELAY-SHARDING](https://github.com/waku-org/specs/blob/master/standards/core/relay-sharding.md)).
```
Please use https://sembr.org/

standards/core/lightpush.md
```diff
@@ -0,0 +1,88 @@
...
+If the node is unable to do so for some reason, they SHOULD return an error code in `LightPushResponse`.
+Once the relay is successful, the `relay_peer_count` will indicate the number of peers that the node has managed to relay the message to. It's important to note that this number may vary depending on the node subscriptions and support for the requested pubsub_topic. The client can use this information to either consider the relay as successful or take further action, such as switching to a lightpush service peer with better connectivity.
```
jm-clius, line 49: Semantic breaks

standards/core/lightpush.md
```diff
@@ -0,0 +1,88 @@
...
+The field `relay_peer_count` may not present or has the value zero in case of error or in other future use cases, where no relay is involved.
```
jm-clius, line 50:
```suggestion
The field `relay_peer_count` may not be present or has the value zero in case of error or in other future use cases, where no relay is involved.
```

jm-clius (approved):
Thanks! I have added some minor comments, but lgtm now. :)

standards/core/lightpush.md
```diff
@@ -0,0 +1,95 @@
...
+---
+
+previous version: `/vac/waku/lightpush/2.0.0-beta1` [19/WAKU2-LIGHTPUSH](https://rfc.vac.dev/waku/standards/core/19/lightpush)
+
+---
+**Protocol identifier**: `/vac/waku/lightpush/3.0.0`
+
+## Motivation and Goals
+
+Light nodes with short connection windows and limited bandwidth wish to push messages to other nodes in the Waku network to request message services.<br>
```
fryorcraken, line 18: instead of `<br>` (`<br />` seems to be the correct tag), just put an empty new line next.

standards/core/lightpush.md
```diff
@@ -0,0 +1,95 @@
...
+A common use case is to request that the service node publish the message to an `11/WAKU2-RELAY` pubsub-topic.
+Additionally, there is sometimes a need for confirmation that a message has been received "by the network"
+(here, at least one node).
+
+`WAKU-LIGHTPUSH` is a request/response protocol for this.
```
fryorcraken, line 23: Are we officially moving from `WAKU2` @jm-clius ?

standards/core/lightpush.md
```diff
@@ -0,0 +1,95 @@
...
+message LightPushRequest {
+    string request_id = 1;
+    // 10 Reserved for future `request_type`. Currently, RELAY is the only available service.
+    optional string pubsub_topic = 20;
+    WakuMessage message = 21;
+}
+
+message LightPushResponse {
+    string request_id = 1;
+    uint32 status_code = 10; // non zero in case of failure, see appendix
```
fryorcraken, line 39: > // non zero in case of failure, see appendix

Yes, `200` is described for success.

standards/core/lightpush.md
```diff
@@ -0,0 +1,95 @@
...
+    optional uint32 relay_peer_count = 12; // number of peers, message is successfully relayed to 
+}
+```
+
+### Message Relaying
+
+Nodes that respond to `LightPushRequest` SHOULD
+- either relay the encapsulated message via [11/WAKU2-RELAY](https://rfc.vac.dev/waku/standards/core/11/relay) protocol on the specified `pubsub_topic`<br>
+- or perform another requested service.
+  `Services beyond [11/WAKU2-RELAY](https://rfc.vac.dev/waku/standards/core/11/relay) are as yet underdefined.`
```
fryorcraken, line 50: > `Services beyond [11/WAKU2-RELAY](https://rfc.vac.dev/waku/standards/core/11/relay) are as yet underdefined.`

simplify: "are yet to be defined"

standards/core/lightpush.md
```diff
@@ -0,0 +1,95 @@
...
+
+Depending on the network configuration, the lightpush client may not need to provide `pubsub_topic` ([WAKU2-RELAY-SHARDING](https://github.com/waku-org/specs/blob/master/standards/core/relay-sharding.md)).<br>
+
+If the node is unable to perform the request for some reason, it SHOULD return an error code in `LightPushResponse`.<br>
+
+Once the relay is successful, the `relay_peer_count` will indicate the number of peers that the node has managed to relay the message to. It's important to note that this number may vary depending on the node subscriptions and support for the requested pubsub_topic. The client can use this information to either consider the relay as successful or take further action, such as switching to a lightpush service peer with better connectivity.<br>
```
fryorcraken, line 56: sembr

standards/core/lightpush.md
```diff
@@ -0,0 +1,95 @@
...
+| BAD_REQUEST | 400   | Wrong request payload.    |
+| PAYLOAD_TOO_LARGE | 413 | Message exceeds certain size limit, it can depend on network configuration, see status_desc for details.  |
+| UNSUPPORTED_PUBSUB_TOPIC | 421 | Requested push on pubsub_topic is not possible as the service node does not support it. |
+| TOO_MANY_REQUESTS | 429 | DOS protection prevented this request as the current request exceeds the configured request rate. |
+| INTERNAL_SERVER_ERROR  | 500 | status_desc holds explanation of the error.  |
+| NO_PEERS_TO_RELAY | 503 | Lightpush service is not available as the node has no relay peers. |
+
+> The list of error codes is not complete and can be extended in the future.
+
+## Security Considerations
```
fryorcraken, line 73: I would add that the content of the response node does not provide guarantees that the messages was relayed and that the requesting node should mitigate appropriately.

standards/core/lightpush.md
```diff
@@ -0,0 +1,95 @@
...
+- DOS protection through request rate limitation on the service itself.
+- message rate limiting via [17/WAKU2-RLN-RELAY](https://rfc.vac.dev/waku/standards/core/17/rln-relay), applied via network membership subscription.
+
+> These features are under development. 
+
+## Future work
+
+- Add support attaching RLN proof for the message requested to be relayed.
+- Add support for other request types.
+- Incentivization of the service
```
fryorcraken, line 86: Definition of strategy to improve reliable usage of this service

NagyZoltanPeter:
@fryorcraken thank you for the comments, let's continue tracking those here: https://github.com/waku-org/specs/pull/29

standards/core/lightpush.md
jm-clius: Since this spec is not numbered yet (a.k.a. it's not a Vac RFC), the title is not critical.
For this to be promoted to Vac RFC, I suggest we keep using `WAKU2` in the official title (also for consistency with other protocols, including the new Store). We can drop the "v2" from the name though - this is simply Waku Lightpush.

The title will be:
`XX/WAKU2-LIGHTPUSH`

where XX is the Vac-assigned RFC number.

PR #22, opened 2024-06-12
feat: waku sync

SionoiS:
First draft of what a Waku Sync specification would look like.

I'm not sure about the direction, do we want more details or maybe write more about what we want in the future?

I feel like it should be really short since it only wraps Negentropy.

WDYT?

### DIFF
--- a/standards/core/sync.md
+++ b/standards/core/sync.md
@@ -7,18 +7,18 @@ contributors:
  - Hanno Cornelius <hanno@status.im>
 ---
 
-# Abstract
+## Abstract
 This specification explains the `WAKU-SYNC` protocol
 which enables the reconciliation of two sets of message hashes
-in the context of keeping multiple Store nodes syncronized.
+in the context of keeping multiple Store nodes synchronized.
 Waku Sync is a wrapper around
 [Negentropy](https://github.com/hoytech/negentropy) a [range-based set reconciliation protocol](https://logperiodic.com/rbsr.html).
 
-# Specification
+## Specification
 
 **Protocol identifier**: `/vac/waku/sync/1.0.0`
 
-## Terminology
+### Terminology
 The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”, “SHOULD”, “SHOULD NOT”, 
 “RECOMMENDED”, “MAY”, and “OPTIONAL” in this document are to be interpreted as described in [RFC2119](https://www.ietf.org/rfc/rfc2119.txt).
 
@@ -28,15 +28,15 @@ the messages created by the Negentropy protocol.
 Client always refers to the initiator
 and the server the receiver of the first payload.
 
-## Design Requirements
+### Design Requirements
 Nodes enabling Waku Sync SHOULD
 manage and keep message hashes in a local cache
 for the range of time
-during which syncronization is required.
+during which synchronization is required.
 Nodes SHOULD use the same time range,
 for Waku we chose one hour as the global default.
 Waku Relay or Light Push protocol MAY be enabled
-and used in conjuction with Sync
+and used in conjunction with Sync
 as a source of new message hashes
 for the cache.
 
@@ -44,7 +44,7 @@ Nodes MAY use the Store protocol
 to request missing messages once reconciliation is complete
 or to provide messages to requesting clients.
 
-## Payload
+### Payload
 
 ```protobuf
 syntax = "proto3";
@@ -58,7 +58,7 @@ message SyncPayload {
 }
 ```
 
-## Session Flow
+### Session Flow
 A client initiate a session with a server
 by sending a `SyncPayload` with
 only the `negentropy` field set to it.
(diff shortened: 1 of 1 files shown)


standards/core/sync.md
```diff
@@ -0,0 +1,93 @@
...
+
+The term Negentropy refers to the protocol of the same name.
+Negentropy payload refers to
+the messages created by the Negentropy protocol.
+Client always refers to the initiator
+and the server the receiver of the first payload.
+
+## Design Requirements
+Nodes enabling Waku Sync SHOULD have the relay and store protocols enabled and
+keep messages for at least the last hour. TODO do we need to say this, sounds like an impl. detail
```
jm-clius, line 33: The reference to an hour is indeed an implementation detail. We want to write "friendly" specs, in that we shouldn't avoid specific examples that may help understanding. I just wouldn't phrase it as part of the protocol.
The requirements are:
- nodes that send and receive Waku messages (most likely via Waku Relay, but they could also be using Waku Filter)
- subset of these nodes that want to cache and synchronise the Waku messages (most likely because they run Waku Store service)

In other words, I wouldn't make the protocol dependent on Relay and Store, but would mention that this is the most obvious example use case. Then I'd say:
- Sync nodes maintain a representative set of message hashes corresponding to locally cached messages
- This representative set should cover the time period over which the nodes require to sync. 1 hour is a sensible default here.
- The sync protocol operates on message hashes and assume that after reconciliation of message hash sets, the full contents of messages can be retrieved in a secondary step (e.g. via Store protocol) and added to the local now-synced cache

In other words:
- describe the function not the protocol
- frame the most likely specific protocols performing functions above as examples
- where implementation is involved, name (but do not mandate) sensible defaults to be as useful as possible

standards/core/sync.md
```diff
@@ -0,0 +1,93 @@
...
+the client computes an empty payload.
+This client computation also outputs any hash differences found,
+those MUST be stored.
+In the case of an empty payload,
+the client MUST send back a `SyncPayload`
+with all the hashes previoudly found in the `hashes` field and
+an empty `engentropy` field.
+
+### Storage Pruning
+TODO? do we need to talk about that, seams like an implementation detail
```
jm-clius, line 73: Perhaps just note under Design Requirements that the size of the local set of message hashes should be managed to correspond to the time period over which you want to sync. No need for negentropy implementation specifics.

jm-clius (commented):
I think this is a good start for a basic spec. I would indeed remain light on details. A separate spec should explain the integrated approach of a synchronised Store (Store-Sync) using resume and Waku-Sync with reference to this spec.
I would avoid tying any directives in this doc directly to any other Waku Protocol, although referencing Relay and Store as the most obvious environment in which Sync can be setup makes sense to me.

SionoiS:
Some of this spec will have to change for 2.0 but lets merge.

standards/core/sync.md
```diff
@@ -0,0 +1,131 @@
...
+
+message SyncPayload {
+  optional bytes negentropy = 1;
+
+  repeated bytes hashes = 20;
+}
+```
+
+### Session Flow
+A client initiate a session with a server
```
jm-clius, line 62:
```suggestion
A client initiates a session with a server
```

standards/core/sync.md
```diff
@@ -0,0 +1,131 @@
...
+by sending a `SyncPayload` with
+only the `negentropy` field set to it.
```
jm-clius, line 64:
```suggestion
only the `negentropy` field set.
```

standards/core/sync.md
```diff
@@ -0,0 +1,131 @@
...
+
+The server receives a `SyncPayload`.
+A new negentropy payload is computed from the received one.
+The server sends back a `SyncPayload` to the client.
+
+The client receives a `SyncPayload`.
+A new negentropy payload OR an empty one is computed.
+If a new payload is computed then
+the exchanges between client and server continues until
+the client computes an empty payload.
```
jm-clius, line 78: What would an empty payload imply? That all hash differences has been exchanged and reconciled?

standards/core/sync.md
```diff
@@ -0,0 +1,131 @@
...
+This client computation also outputs any hash differences found,
+those MUST be stored.
+In the case of an empty payload,
+the client MUST send back a `SyncPayload`
+with all the hashes previously found in the `hashes` field and
```
jm-clius, line 83: Not sure I get the formulation here. What does "hashes previously found" mean? All the hashes that the client discovered to be missing in its own storage or those it has found locally but not in the server's set?

jm-clius (approved):
Approving as a raw spec. I think this will change substantially with the intro of Sync 2.0.

standards/core/sync.md
SionoiS: Yes exactly

standards/core/sync.md
SionoiS: Yeah quite unclear. It's the hashes that the server is missing...

PR #23, opened 2024-06-25
feat: reliability for relay protocol

kaichaosun:
TODOs,
- [x] implementations

### DIFF
--- a/standards/application/relay-reliability.md
+++ b/standards/application/relay-reliability.md
@@ -106,11 +106,10 @@ An application could use different pubsub topics and content topics, for example
 
 For incoming messages, the processing flow can be like this:
 - subscribe to the interested pubsub and content topics
-- query the store node with the interested topics and time range for message hashes periodically
+- periodically query the store node with the interested topics and time range for message hashes
 - check if each received message hash already exists in the local database. if not, add the missing message hash to a buffer.
 - batch fetch the full messages corresponding to the missing message hashes in the buffer from the store node
 - process the messages
-- update the last fetch time for the interested topic
 
 The implementation in Python may look like this:
 


standards/application/relay-reliability.md
```diff
@@ -0,0 +1,150 @@
...
+  - Richard Ramos <richard@status.im>
+---
+
+## Abstract
+
+[Relay Protocol](https://github.com/vacp2p/rfc-index/blob/main/waku/standards/core/11/relay.md) in Waku is efficient for routing messages, but there's no guarantee that a message will reach to its destination. For example, the receiver in a chat application may miss messages when network issue happens at either sender or receiver side. In general, a message in Waku network includes 3 status:
+
+- **outgoing**, the message is posted by its creator but no confirmations from other nodes yet
+- **sent**, the message is received by any other node in the network
+- **delivered**, the message is acknowledged by the receiver
```
jm-clius, line 17: This is true of applications, but is "delivered" something that really has meaning on the Waku layer? Waku is a broadcast protocol - in other words, the aim is to deliver the message "to everyone" (with the application determining the intended recipient). To me it seems like in such a system you only have "send attempted" and "send acknowledged" with the latter being a variable parameter in terms of how many nodes in the broadcast group have acknowledged your message. This is also only a status that makes sense from the publisher's point of view.

standards/application/relay-reliability.md
```diff
@@ -0,0 +1,150 @@
...
+
+Application like Status already uses [MVDS](https://github.com/vacp2p/rfc-index/blob/main/vac/2/mvds.md) for e2e acknowledgement in direct messages and group chat. There is an ongoing [discussion](https://forum.vac.dev/t/end-to-end-reliability-for-scalable-distributed-logs/293) about a more general and bandwidth efficient solution for e2e reliablity.
```
jm-clius, line 19: Can we be a bit more explicit about what constitutes reliability from application vs Waku POV. In other words, an application defines a payload over Waku and is interested in e2e delivery between application users. Waku is a pub/sub broadcast transport interested in reliably routing a message to all participants in the pub/sub broadcast group.

standards/application/relay-reliability.md
```diff
@@ -0,0 +1,150 @@
...
+
+The [Store protocol](https://github.com/waku-org/specs/blob/master/standards/core/store.md) provides a way for nodes in the network to query the existence of messages or fetch specific messages based on the search criteria.
+
+**Search criteria with message hash**
+
+For the nodes that may have connection issues to **publish** messages via relay network, this search criteria can be used to check whether a message is populated in the network or not. The message exists in store node can be marked from `outgoing` to `sent` by application. If the message is not found in the store node, the application can resend the message.
+
+**Search criteria with topics and time range**
+
+For the nodes that may have connection issues to **receive** messages via relay network, this search criteria can be used to fetch missing messages from store nodes when network resumes.
```
jm-clius, line 33: "when network resumes"

As far as I understand the implementation further down, this is not triggered only when a connection resumes but periodically (in order to also cover undetected "offline" periods)?

standards/application/relay-reliability.md
```diff
@@ -0,0 +1,150 @@
...
+
+For the nodes that may have connection issues to **publish** messages via relay network, this search criteria can be used to check whether a message is populated in the network or not. The message exists in store node can be marked from `outgoing` to `sent` by application. If the message is not found in the store node, the application can resend the message.
+
+**Search criteria with topics and time range**
+
+For the nodes that may have connection issues to **receive** messages via relay network, this search criteria can be used to fetch missing messages from store nodes when network resumes. 
+
+By leveraging the store node to provide such query services, the applications are able to mitigate the reliability issue of relay protocol. But this approach also introduces new challenges like centralization, privacy, and scalability. It should be viewed as a temporary solution and deprecated when e2e reliability solution is ready.
+
+## Implementation Suggestions
```
jm-clius, line 37: An Implementation Suggestion can be added towards the end of the spec, but I'd describe the mechanism here in a simple heuristic. Also, since we already do periodic store queries to find missing received messages, I don't see the need for the "queries with message hash" for outgoing messages. Why not simply verify that your outgoing message hashes are included in the periodic query results? This way we can have a consolidated mechanism that deals with both outgoing/publish and incoming/received message reliability.

In fact, I think the heuristic will be simpler to describe than an implementation suggestion. Very roughly:

1. create outgoing buffer for all "outgoing" messages
2. create incoming buffer for all recently received message hashes
3. periodic store query based on time range, returning only message hashes
4. check outgoing buffer: message hashes included in response from (3) are "sent", others are resent
5. check incoming buffer: message hashes included in response from (3) that's not in incoming buffer are missing. Request these from the store.

There will of course be more details, but I think the main focus of this spec should be a simple, consolidated mechanism close to the above. WDYT?

standards/application/relay-reliability.md
```diff
@@ -0,0 +1,150 @@
...
+    response.messages().forEach(message -> {
+        processMessage(message)
+    })
+
+    updateFetchRecord(queryParams.pubsubTopic, queryParams.toTime)
+```
+
+`QueryParam` includes all the necessary information to fetch missing messages, it may also include peer id of the store node for easy discovery. The application should iterate all the interested pubsub topics, along with its content topics to construct the `QueryParam`.
+
+Function `fetchMissingMessages` is runing periocally, for example 1 minute. It first fetch all the message hashes in the specified time range, check if message exist in local dabatase, if not, fetch the missing messages in batch. The batch size should be bounded to avoid large data transfer or exceed the max supported size by store node.
```
jm-clius, line 124: When we make the simplifying assumption that nodes publish and subscribe to the same pubsub topic/content topics, this periodic store query response can also be used to check if outgoing messages have indeed been sent. In the case of most deployments (including Status) this assumption holds true, so I think we can focus on such a simplified single periodic query algorithm.

jm-clius (changes requested):
Thanks a lot for starting to clarify these convoluted concepts! I have added some suggestions below which I think will help us:
(a) reason a bit more generally about Waku reliability (and not just with a Relay focus)
(b) consolidate/simplify the strategies for outgoing and incoming reliability

Let me know if my comments are not clear. :)

standards/application/relay-reliability.md
richard-ramos: Yes, this is executed periodically regardless of the network connectivity state

standards/application/relay-reliability.md
```diff
@@ -0,0 +1,150 @@
...
+                continue
+            }
+            missingMessageHashes.append(messageHash)
+        })
+
+        # process next page of the response
+        response.Next()
+    
+    # fetch missing messages with hashes in batch
+    response = waku.store.queryMessages(queryParams, missingMessageHashes)
```
richard-ramos, line 114: `queryParams` is not needed here, as `missingMessageHashes` will contain the primary keys for the table (it is also not supported). I also renamed the function to `queryMessagesByHash` to differentiate it from the other function as well as added a `peerID`

```suggestion
    response = waku.store.queryMessagesByHash(peerID, missingMessageHashes)
```

standards/application/relay-reliability.md
```diff
@@ -0,0 +1,150 @@
...
+    pubsubTopic: str
+    contentTopics: List[str]
+    fromTime: int
+    toTime: int
+
+def fetchMissingMessages(queryParams):
+    missingMessageHashes = []
+
+    # get missing message identifiers first in order to reduce the data transfer
+    response = waku.store.queryMessageHashes(queryParams)
```
richard-ramos, line 99:
```suggestion
    response = waku.store.queryMessageHashes(peerID, queryParams)
```

standards/application/relay-reliability.md
```diff
@@ -0,0 +1,150 @@
...
+    # fetch missing messages with hashes in batch
+    response = waku.store.queryMessages(queryParams, missingMessageHashes)
+    response.messages().forEach(message -> {
+        processMessage(message)
+    })
+
+    updateFetchRecord(queryParams.pubsubTopic, queryParams.toTime)
+```
+
+`QueryParam` includes all the necessary information to fetch missing messages, it may also include peer id of the store node for easy discovery. The application should iterate all the interested pubsub topics, along with its content topics to construct the `QueryParam`.
```
richard-ramos, line 122: `QueryParam` as defined in line 89, does not have the peerID. I added it as a parameter to `queryMessages`

richard-ramos (approved):

standards/application/relay-reliability.md
```diff
@@ -0,0 +1,150 @@
+---
+title: RELIABILITY-RELAY
+name: Reliability for Relay Protocol
```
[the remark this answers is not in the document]
jm-clius: > I'm not sure if we are can treat lightpush/filter as another Transport Layer of Waku

Indeed. I think this is the point - if we abstract Waku as a p2p communictions transport layer (regardless of its own individual hop transports), lightpush and filter are just two wrappers that allow access to this communications layer. Store-based reliability is built (or at least defined) on top of this with no specific requirements on how this p2p routing/transport layer is accessed (directly via relay or indirectly via filter/lightpush). The reason we need the specification to be more general is so we can implement this for light clients too - I believe regular store queries will be necessary to implement lightpush/filter reliably too.

>  should be the target for our e2e reliability proposal

Not sure I agree. Perhaps the confusion might be with the word "transport" and we can perhaps better substitute it for a term like "routing". Waku Transport is the set of Waku protocols that allow delivery of anonymous messages in a p2p broadcast configuration (with no insight into the payload). I'm not referring here to the underlying transports on which the individual hops are built (which may indeed use Tor). This spec is concerned with the reliability of this routing mechanism, combined as a new p2p transport layer on which applications such as Status can build.

E2E reliability is concerned with the reliable delivery from an application's perspective (i.e. with insight into the payload and intended recipients).

standards/application/relay-reliability.md
jm-clius: >  The states and terms used here is trying to mimic from app's perspective, so it can be better understandable by app devs.

It is OK to have a transport spec contextualise itself in terms that make sense for typical app developers, but it needs to be explicit about separation of concerns here. Store queries and responses on their own cannot know if an app's messages are "delivered".

> state changes from receiver side

Indeed. My point is simply to make it explicit that we're referring to state from sender POV.

standards/application/relay-reliability.md
jm-clius: > to do such refactor in status-go is not a trivial work

Valid concern, though I would focus on describing the concept as simply (and puristically) as possible in the spec even if our initial implementation(s) may not fully implement the spec. In other words, I wouldn't write a spec based on the limitations of the implementation, but extract the concepts in such a way that we can, incrementally, bring the implementation to fully implement the spec even if for now we start with what is trivial to add.

> the requirement is different

This is indeed a very good point. In a heuristic like above, this would add an optional mechanism for more regular (and even more cheap) outgoing message checks.

standards/application/relay-reliability.md
```diff
@@ -0,0 +1,192 @@
...
+category: Standards Track
+tags: [reliability, application]
+editor: Kaichao Sun <kaichao@status.im>
+contributors:
+  - Richard Ramos <richard@status.im>
+---
+
+## Abstract
+
+Waku is efficient for routing messages with its transport layer like [Relay protocol](https://github.com/vacp2p/rfc-index/blob/main/waku/standards/core/11/relay.md) and [Lightpush](https://github.com/vacp2p/rfc-index/blob/main/waku/standards/core/19/lightpush.md) / [Filter](https://github.com/vacp2p/rfc-index/blob/main/waku/standards/core/12/filter.md) protocol. But there's no guarantee that a message going through Waku will reach to its destination. For example, the receiver in a chat application may miss messages when network issue happens at either sender or receiver side.
```
jm-clius, line 13: A bit nitpicky, but I would rephrase to make it clear that Waku Relay (with Filter/Lightpush) _is_ a p2p transport layer. If we phrase it like the transport layer just belongs to Waku, it may be confusing because the Waku Transport does indeed also build on hop-to-hop transport layers like QUIC (not implemented yet), TCP, websockets, etc.

jm-clius, line 13: Please add more linebreaks in accordance with sembr.

jm-clius, line 13: As a suggestion for clarity:

```suggestion
Waku provides an efficient transport layer for p2p communications.
It defines protocols like [Relay](https://github.com/vacp2p/rfc-index/blob/main/waku/standards/core/11/relay.md) and [Lightpush](https://github.com/vacp2p/rfc-index/blob/main/waku/standards/core/19/lightpush.md) / [Filter](https://github.com/vacp2p/rfc-index/blob/main/waku/standards/core/12/filter.md) for routing messages in decentralised networks.
However, there is no guarantee that a message broadcast in a Waku network will reach its destination.
For example, the receiver in a chat application using Waku as p2p transport may miss messages
when a network issue happens at either the sender or the receiver side. 
```

standards/application/relay-reliability.md
```diff
@@ -0,0 +1,192 @@
...
+editor: Kaichao Sun <kaichao@status.im>
+contributors:
+  - Richard Ramos <richard@status.im>
+---
+
+## Abstract
+
+Waku is efficient for routing messages with its transport layer like [Relay protocol](https://github.com/vacp2p/rfc-index/blob/main/waku/standards/core/11/relay.md) and [Lightpush](https://github.com/vacp2p/rfc-index/blob/main/waku/standards/core/19/lightpush.md) / [Filter](https://github.com/vacp2p/rfc-index/blob/main/waku/standards/core/12/filter.md) protocol. But there's no guarantee that a message going through Waku will reach to its destination. For example, the receiver in a chat application may miss messages when network issue happens at either sender or receiver side. 
+
+In general, a message in Waku network includes 3 status from sender's perspective:
```
jm-clius, line 15:
```suggestion
In general, a message in a Waku network may be in one of 3 states from the sender's perspective:
```

standards/application/relay-reliability.md
```diff
@@ -0,0 +1,192 @@
...
+
+- **outgoing**, the message is posted by its creator but no confirmations from other nodes yet
```
jm-clius, line 17: I would define either sender/publisher (or both) and try to stick to the same terminology throughout. "creator" is a new term here and I think can be consistent in using "publisher" here.

standards/application/relay-reliability.md
jm-clius: I still think "delivered" here changes the agents suddenly from sender/receiver to sending application instance and receiving application instance. Perhaps OK to just phrase a bit more clearly: "the message is acknowledged on the application layer by the intended recipient"

standards/application/relay-reliability.md
```diff
@@ -0,0 +1,192 @@
...
+- **sent**, the message is received by any other node in the network
+- **delivered**, the message is acknowledged by the receiver
+
+Application like Status already uses [MVDS](https://github.com/vacp2p/rfc-index/blob/main/vac/2/mvds.md) for e2e acknowledgement in direct messages and group chat. There is an ongoing [discussion](https://forum.vac.dev/t/end-to-end-reliability-for-scalable-distributed-logs/293) about a more general and bandwidth efficient solution for e2e reliablity.
```
jm-clius, line 21: sembr, please

standards/application/relay-reliability.md
```diff
@@ -0,0 +1,192 @@
...
+
+In other words, an application defines a payload over Waku and is interested in e2e delivery between application users. Waku provides a pub/sub broadcast transport, which is interested in reliably routing a message to all participants in the broadcast group.
+
+Before we have a complete design for e2e reliability, we need to compose existing protocols to increase the reliability of the transport protocol. This document proposes a few options for such composition. 
+
+## Motivation
+
+The [Store protocol](https://github.com/waku-org/specs/blob/master/standards/core/store.md) provides a way for nodes in the network to query the existence of messages or fetch specific messages based on the search criteria.
+
+**Search criteria with message hash**
```
jm-clius, line 31:
```suggestion
**Query criteria with message hash**
```

jm-clius: Here and throughout I'd use "query" instead of search in order to limit the number of terms we introduce.

standards/application/relay-reliability.md
```diff
@@ -0,0 +1,192 @@
...
+
+For the nodes that may have connection issues to **publish** messages via transport layer, this search criteria can be used to check whether a message is populated in the network or not. The message exists in store node can be marked from `outgoing` to `sent` by application. If the message is not found in the store node, the application can resend the message.
```
jm-clius, line 33: sembr, please (here and throughout)

standards/application/relay-reliability.md
```diff
@@ -0,0 +1,192 @@
...
+
+**Search criteria with topics and time range**
+
+For the nodes that may have connection issues to **receive** messages via transport layer, this search criteria can be used to fetch missing messages from store nodes periodically after network resumes. 
+
+In summary, by leveraging the store node to provide such query services, the applications are able to mitigate the reliability issue of transport layer. This approach also introduces new challenges like centralization, privacy, and scalability. It should be viewed as a temporary solution and deprecated when e2e reliability solution is ready.
```
jm-clius, line 39:
```suggestion
In summary, by leveraging the store node to provide such query services, the applications are able to mitigate reliability issues on the Waku transport layer.
This approach also introduces new limitations like centralized points of failures in Store nodes, diminished privacy, and lower scalability.
It should be viewed as a temporary solution and deprecated when an e2e reliability solution is ready.
```

standards/application/relay-reliability.md
```diff
@@ -0,0 +1,192 @@
...
+
+## Implementation Suggestions 
+
+### Query with Message Hash
+
+For outgoing messages, the processing flow can be like this:
+- create a buffer for all "outgoing" messages
```
jm-clius, line 46:
```suggestion
- create a buffer for all "outgoing" message hashes
```
I think? We should be explicit about what goes in the buffer, what the role of the "local database" is, etc.

standards/application/relay-reliability.md
```diff
@@ -0,0 +1,192 @@
...
+- send message via relay or lightpush protocol
+- add message hash to the buffer
+- save the message to local database with status "outgoing"
```
jm-clius, line 49: I'd avoid introducing a non-protocol related concept like "database" here. We can just say "keep a copy of the message locally with status outgoing".

standards/application/relay-reliability.md
```diff
@@ -0,0 +1,192 @@
...
+- check the buffer periodically
+- query the store node with message hash in the buffer, the messages should be posted more than a few seconds ago
+- if the message exists, update the status to "sent" and remove the message hash from the buffer
```
jm-clius, line 52: update the status to "sent" of the local copy of the message, I assume?

standards/application/relay-reliability.md
```diff
@@ -0,0 +1,192 @@
...
+
+### Query with Message Hash
+
+For outgoing messages, the processing flow can be like this:
+- create a buffer for all "outgoing" messages
+- send message via relay or lightpush protocol
+- add message hash to the buffer
+- save the message to local database with status "outgoing"
+- check the buffer periodically
+- query the store node with message hash in the buffer, the messages should be posted more than a few seconds ago
```
jm-clius, line 51: query the store node for each message hash in the buffer of which the send attempt was more than a few seconds ago

standards/application/relay-reliability.md
```diff
@@ -0,0 +1,192 @@
...
+- if the message exists, update the status to "sent" and remove the message hash from the buffer
+- if the message does not exist, resend the message
+- if the message is still missing in the store node for a period of time, trigger the message failed to send workflow and remove the message hash from the buffer
+
+The implementation in Python may look like this:
```
jm-clius, line 56: I would still move all the implementation suggestions to a separate section at the end of the spec. The heuristics can then belong to a single "Protocol" subheader and have subheadings that are more descriptive, e.g.

```
## Protocol

### Outgoing message reliability

### Incoming message reliability

### Unified approach

## Implementation suggestion

### Query with Message Hash

### Query with Topics and Time Range

```

standards/application/relay-reliability.md
```diff
@@ -0,0 +1,192 @@
...
+When persistent network issue happens, you may not want to resend the failed messages indefinitely, the application should have a mechanism to clean the cache with failed message hashes and trigger other retry logic after a few attempts.
+
+### Query with Topics and Time Range
+
+An application could use different pubsub topics and content topics, for example a community may have its own pubsub topic, and each channel may have its own content topic. To fetch all missing messages in a specific channel, the application can query the store node with the provided pubsub topic, content topic and time range.
+
+For incoming messages, the processing flow can be like this:
+- subscribe to the interested pubsub and content topics
+- query the store node with the interested topics and time range for message hashes periodically
+- check if the message hash is exist in local database, if not, adding the message hash to a buffer, if yes, skip the message
```
jm-clius, line 104:
```suggestion
- check if each received message hash already exists in the local database. if not, add the missing message hash to a buffer.
```

standards/application/relay-reliability.md
```diff
@@ -0,0 +1,192 @@
...
+- fetch the missing messages in the buffer from store node
```
jm-clius, line 105:
```suggestion
- batch fetch the full messages corresponding to the missing message hashes in the buffer from the store node
```

standards/application/relay-reliability.md
```diff
@@ -0,0 +1,192 @@
...
+- process the messages necessarily
```
jm-clius, line 106: Not sure what "necessarily" means in this context? Sequentially?

standards/application/relay-reliability.md
```diff
@@ -0,0 +1,192 @@
...
+- update the last fetch time for the interested topic
```
jm-clius, line 107: I think this can be removed from the heuristic if we just make it clear that the heuristic must be run periodically for all content topics interesting to the appplication.

jm-clius (approved):
Thanks a lot! Approving, as I think this is a good start for a raw protocol and can be merged and iterated upon. I have added some comments below, but the core of the spec seems good.

PR #24, opened 2024-06-25
Fix Broken Links

jimstir:
Specification linked to old rfc.vac.dev/spec format. Updated links to new format.


jm-clius (approved):
Thanks!

PR #28, opened 2024-07-23
chore: clarifications for node health when disparity in protocol health

danisharora099:
## Problem

When talking about the Node health, it is categorised into two buckets:
- protocol health
- overall node health

When all the protocols report the same health metric, it's straightforward that the node health follows the same result.

In case where individual protocols report different health, we don't specify our recommendation.

## Suggested Solution

If there is a disparity in the node health for each protocol, we SHOULD derive the node health from the worse of the two.
Example: if LightPush has 2 peers (sufficiently healthy), but Filter has 0 available peers (unhealthy), overall node health should be reported as _unhealthy_

### DIFF
--- a/standards/application/req-res-reliability.md
+++ b/standards/application/req-res-reliability.md
@@ -47,7 +47,7 @@ To address this we RECOMMEND following states:
 - sufficiently healthy:
   - Filter has at least 2 connections available to service nodes AND LightPush has at least 2 connections available to service nodes;
 
-If one of the protocols is unhealthy or minimally healthy, while the other is sufficiently healthy - overall node health SHOULD be considered as unhealthy or minimally healthy respectively.
+Overall node health SHOULD be considered as unhealthy or minimally healthy respectively if one of the protocols is unhealthy or minimally healthy, while the other is sufficiently healthy.
 
 ### Peers and connection management
 


standards/application/req-res-reliability.md
```diff
@@ -40,13 +40,14 @@ Node health is a metric meant to determine the connectivity state of a light nod
 We consider this reliability to be dependent on amount of simultaneous connections to responsive service nodes.
 Unfortunately the more connections light node establishes - the more bandwidth is consumed.
 To address this we RECOMMEND following states:
-- unhealthy - no connections to service nodes are available regardless of protocol;
+- unhealthy:
+  - no connections to service nodes are available regardless of protocol.
```
weboko, line 44:
```suggestion
  - no connections to service nodes are available regardless of protocol;
```

standards/application/req-res-reliability.md
```diff
@@ -40,13 +40,14 @@ Node health is a metric meant to determine the connectivity state of a light nod
...
 - minimally healthy:
-  - Filter has one service node connection;
-  - LightPush protocol has one service node connection;
+  - Filter has one service node connection AND LightPush protocol has one service node connection.
```
weboko, line 46:
```suggestion
  - Filter has one service node connection AND LightPush protocol has one service node connection;
```

standards/application/req-res-reliability.md
```diff
@@ -40,13 +40,14 @@ Node health is a metric meant to determine the connectivity state of a light nod
...
 - sufficiently healthy:
-  - Filter has at least 2 connections available to service nodes;
-  - LightPush has at least 2 connections available to service nodes;
+  - Filter has at least 2 connections available to service nodes AND LightPush has at least 2 connections available to service nodes.
```
weboko, line 48:
```suggestion
  - Filter has at least 2 connections available to service nodes AND LightPush has at least 2 connections available to service nodes;
```

standards/application/req-res-reliability.md
```diff
@@ -40,13 +40,14 @@ Node health is a metric meant to determine the connectivity state of a light nod
...
+
+If one of the protocols is unhealthy or minimally healthy, while the other is sufficiently healthy - overall node health SHOULD be considered as unhealthy or minimally healthy respectively.
```
weboko, line 50: I would place recommendation first

```suggestion
Overall node health SHOULD be considered as unhealthy or minimally healthy - if one of the protocols is unhealthy or minimally healthy, while the other is sufficiently healthy.
```

weboko (approved):
thanks for keeping it up to date!

PR #34, opened 2024-08-30
Refinements of RLN on mainnet spec

s-tikhomirov:
This PR contains relateively minor edits to the RLN contract spec that resulted from discussion during its implementation.

### DIFF
--- a/standards/core/rln-contract.md
+++ b/standards/core/rln-contract.md
@@ -53,18 +53,21 @@ and MAY also use additional criteria.
 
 Contract parameters and their RECOMMENDED values for the initial mainnet deployment are as follows:
 
-| Parameter                                               | Symbol    | Value    | Units                |
-| ------------------------------------------------------- | --------- | -------- | -------------------- |
-| Epoch length                                            | `epoch`   | `10`     | minutes              |
-| Maximum total rate limit of all memberships in the tree | `R_{max}` | `160000` | messages per `epoch` |
-| Minimum rate limit of one membership                    | `r_{min}` | `20`     | messages per `epoch` |
-| Maximum rate limit of one membership                    | `r_{max}` | `600`    | messages per `epoch` |
-| Membership price for `1` message per epoch              | `p_u`     | `0.05`   | `USD`                |
-| Membership expiration term                              | `T`       | `180`    | days                 |
-| Membership grace period                                 | `G`       | `30`     | days                 |
-| Accepted tokens                                         |           | `DAI`    |                      |
+| Parameter                                               | Symbol    | Value    | Units              |
+| ------------------------------------------------------- | --------- | -------- | ------------------ |
+| Epoch length                                            | `t_{ep}`  | `600`    | seconds            |
+| Maximum total rate limit of all memberships in the tree | `R_{max}` | `160000` | messages per epoch |
+| Minimum rate limit of one membership                    | `r_{min}` | `20`     | messages per epoch |
+| Maximum rate limit of one membership                    | `r_{max}` | `600`    | messages per epoch |
+| Membership price for `1` message per epoch              | `p_u`     | `0.05`   | `USD`              |
+| Membership expiration term                              | `T`       | `180`    | days               |
+| Membership grace period                                 | `G`       | `30`     | days               |
+| Accepted tokens                                         |           | `DAI`    |                    |
 
 The pricing function SHOULD be linear in the rate limit per epoch.
+
+Note: epoch length means the same as `period` as defined in [17/WAKU2-RLN-RELAY](https://github.com/vacp2p/rfc-index/blob/main/waku/standards/core/17/rln-relay.md).
+This specification uses the term "epoch length" instead of "period" to avoid confusion with "grace period".
 
 ## Membership lifecycle
 


standards/core/rln-contract.md
```diff
@@ -45,27 +45,37 @@ The contract MUST provide the following functionalities:
...
-| Membership grace period                                 | `G`       | `30`     | days                 |
-| Accepted tokens                                         |           | `DAI`    |                      |
+| Parameter                                               | Symbol    | Value    | Units              |
+| ------------------------------------------------------- | --------- | -------- | ------------------ |
+| Epoch length                                            | `t_{ep}`  | `600`    | seconds            |
+| Maximum total rate limit of all memberships in the tree | `R_{max}` | `160000` | messages per epoch |
+| Minimum rate limit of one membership                    | `r_{min}` | `20`     | messages per epoch |
+| Maximum rate limit of one membership                    | `r_{max}` | `600`    | messages per epoch |
+| Membership price for `1` message per epoch              | `p_u`     | `0.05`   | `USD`              |
+| Membership expiration term                              | `T`       | `180`    | days               |
```
alrevuelta, line 70: I would suggest renaming `Membership expiration term` to `Membership billing period` or something similar + slightly changing its definition.

As it is, it gives the impression that your membership will expire after `Membership expiration term` aka `T` aka `180days`. However, I think we should allow the user to register for several *billing periods*.

This can actually allow lowering the `T` from `180days` to eg `30days` which is a more common period (rent, salary, netflix) + it will be less pricy for a user wanting to test the protocol, without having to commit for the price of 6 months. But if the user wishes, he can also register for an arbitrary number of periods.

I would also change the:
* `Membership price for 1 message per epoch` aka `p_u`
* to `Membership price for 1 message per epoch per period` aka `p_u`

And as an implementation detail cc @richard-ramos allow membership registration for an arbitrary number of periods. Where price is calculated as `num_periods * p_u * num_messages`.


In other words, I think we should:
* Allow registering for n amount of periods.
* Perhaps lower 180, since `180*n` doesn't give much granularity

richard-ramos: For grace period, we can either multiply it for the number of periods, (or convert it to a %).
For example, assuming the original value of expiration term of 180 days and 30d grace period, if we change it to periods of 30d, the equivalent grace period would be ~5d, so anyone registering for 6 periods would end up with 30 of grace period (5d*6).

richard-ramos: Should we have a maximum number of periods an user could register as a way of preventing squatting of the slots on the merkle tree or is this a non-issue?

richard-ramos: Another question is,  how would extending a membership work in this case? should they also specify the number of periods they want to extend?

alrevuelta: > Should we have a maximum number of periods an user could register as a way of preventing squatting of the slots on the merkle tree or is this a non-issue?

yes, i think so. eg max 1 or 2 years. this limit is indeed necessary because if someone registers for 10000 years, thats a liability for waku, since waku would need to provide services for 10000 years?

> Another question is, how would extending a membership work in this case? should they also specify the number of periods they want to extend?

yes.

fryorcraken: I would keep a max to 6 months as we are still early stage.

s-tikhomirov: As discussed on research call 2024-09-09, I've created an [issue](https://github.com/waku-org/specs/issues/37) in the spec repo and am resolving this discussion for now. Let's consider this idea more thoroughly for the next iteration of the spec.

> renaming Membership expiration term to Membership billing period or something similar

I'm also adding this terminology clarification to this PR.

standards/core/rln-contract.md
```diff
@@ -127,22 +137,32 @@ Availability of membership-specific functionalities[^2] MUST be as follows:
...
+Let us define the following aggregate rate limits:
+- `R_{active}` is the total rate limit of all _Active_ memberships;
+- `R_{grace_period}` is the total rate limit of all _GracePeriod_ memberships;
+- `R_{expired}` is the total rate limit of all _Expired_ memberships;
+- `r` is the requested rate limit of a new membership.
+
+Let `R_{free} = R_{max} - R_{active} - R_{grace_period} - R_{expired}` be the free rate limit that is available without overwriting _Expired_ slots.
+
+Membership registration is subject to the following requirements:
+- If `r <= R_{free}`, the new membership MUST be registered (assuming all other necessary conditions hold). The new membership MAY overwrite one or multiple _Expired_ memberships.
```
fryorcraken, line 149: > multiple _Expired_ memberships

Good idea.

standards/core/rln-contract.md
```diff
@@ -180,6 +204,17 @@ and the membership set SHOULD be migrated.
...
 ## Implementation Suggestions
 
+### Choosing Which _Expired_ Memberships to Overwrite
+
+When registering a new membership, the smart contract may need to decide which _Expired_ memberships to overwrite.
+The criteria for this selection can vary depending on the implementation.
+
+Key considerations include:
+- To minimize gas costs, it's better to overwrite a single high-rate membership rather than multiple low-rate ones.
+- To encourage timely withdrawals, it's better to overwrite memberships that have been _Expired_ for a long time.
```
fryorcraken, line 214: Do we agree @richard-ramos that for now, we don't include this feature in the smart contract?

richard-ramos: The way it works right now is that you can:
- Specify the memberships you want to expire
- If you don't specify:
- - If there's enough space in the tree / enough rate limit: create the membership
- - Otherwise, attempt to free it.

Since for any option ideally the client should call `estimateGas` to determine which option is cheaper, and the smart contract's been already modified to support this scenario, we can keep it. Unless we want to remove the tracking of the membership age in the smart contract completely? If so, then, cool. It means less gas for the user, and nwaku can track the age of the memberships then.

fryorcraken: > nwaku can track the age of the membership

Are you saying that nwaku would need to listen to emit events from the smart contract? I think we definitely do not want that as the whole point of onchain tree was to not have to load all events.

However, can the solution be:
- membership and some age information are stored in contract
- BUT they are not ordered

Meaning that it's cheaper to insert or expire (no ordering at insertion). But it's possible to get all data from the contract API (no tracing of emitted events).
The ordering can then be done by the applicaiton (nwaku, jswaku) etc

richard-ramos: Sure! no need to use events. There's already a function that would let you iterate over all the memberships registered. Also perhaps in the future we might want to integrate with the graph and obtain this information from there too. 

I'll propose the changes on a separate branch from https://github.com/waku-org/waku-rlnv2-contract/pull/13 so it's possible to compare both options. I do think not having to track the ordering in the contract will be very beneficial due to reduced costs

standards/core/rln-contract.md
```diff
@@ -115,34 +134,50 @@ A user MAY use one Waku node[^1] to manage multiple memberships.
...
+```
+R_{free} = R_{max} - R_{active} - R_{grace_period} - R_{expired}
+```
+
+Membership registration is additionally subject to the following requirements:
+- If `r <= R_{free}`, the new membership MUST be registered (assuming all other necessary conditions hold). The new membership MAY overwrite one or multiple _Expired_ memberships.
+- If `r > R_{free}`:
+	- if `r > R_{free} + R_{expired}`, registration MUST fail;
+	- if `r <= R_{free} + R_{expired}`, the new membership MUST be registered by overwriting some _Expired_ memberships.
+- The sender of the registration transaction MAY specify a list of _Expired_ memberships to be overwritten. If the list is not provided, the contract MAY use any criteria to select _Expired_ memberships to overwrite (see Implementation Suggestions).
```
fryorcraken, line 176: sembr please.

> If the list is not provided, the contract MAY use any criteria to select _Expired_ memberships to overwrite (see Implementation Suggestions).

I would not bother implementing that @richard-ramos

standards/core/rln-contract.md
```diff
@@ -180,12 +216,27 @@ and the membership set SHOULD be migrated.
...
+The membership set MAY be implemented as a Merkle tree, such as an [Incremental Merkle Tree](https://zkkit.pse.dev/modules/_zk_kit_imt.html) (IMT).
+
+### Choosing Which _Expired_ Memberships to Overwrite
+
+When registering a new membership, the contract needs to decide which _Expired_ memberships, if any, to overwrite.
+The criteria for this selection can vary depending on the implementation.
+
+Key considerations include:
+- To minimize gas costs, it's better to overwrite a single high-rate membership rather than multiple low-rate ones.
+- To encourage timely deposit withdrawals, it's better to overwrite memberships that have been _Expired_ for a long time.
```
fryorcraken, line 230: I think this is unneeded complexity in the smart contract. I did push in favour of lazy override initially. But I was convinced otherwise by higher gas cost and fluctuation of price.

IMO it would make more sense for the app to select the membership to override.
Meaning the contract would just revert if the total rate limit is exceed and no memberships are being overridden (no gas lost, right @richard-ramos ?)
It's then up to the application to be clever on how to select memberships to override.

s-tikhomirov:
This PR is ready for review. It corresponds to my other suggested PR to the spec: https://github.com/waku-org/waku-rlnv2-contract/pull/17

richard-ramos (approved):

PR #36, opened 2024-09-06
feat: store sync spec

SionoiS:
New Store sync spec that tie Store Resume and Waku Sync into one strategy/best practice spec.

WIP

### DIFF
--- a/standards/core/store-sync.md
+++ b/standards/core/store-sync.md
@@ -1,13 +1,13 @@
 ---
 title: WAKU-STORE-SYNC
-name: Waku Store Syncronization
+name: Waku Store Synchronization
 editor: Simon-Pierre Vivier <simvivier@status.im>
 contributors:
 ---
 
 ## Abstract
 
-This document describe the strategy Waku store node will employ to stay synchronized.
+This document describe the strategy Waku Store node will employ to stay synchronized.
 The goal being that all store nodes eventually archive the same set of messages.
 
 ## Background / Rationale / Motivation
@@ -27,7 +27,7 @@ Various protocols and features that help with message consistency are described 
 
 ### Store Resume
 
-This feature allow a node to fill the gap in messages for the period it was offline.
+This feature allow a node to fill the gap in messages for the period it was last offline.
 At startup, a node use the Store protocol to query a random node for
 the time interval since it was last online.
 It is RECOMMENDED” to limit the time interval to a maximum of 6 hours.
@@ -35,7 +35,7 @@ It is RECOMMENDED” to limit the time interval to a maximum of 6 hours.
 ### Waku Sync
 
 Nodes that stay online can still miss messages.
-Waku Sync is used to find those messages by periodically syncing with random node.
+Waku Sync is used to find those messages by periodically syncing with random nodes.
 It is RECOMMENDED to trigger a sync every 5 minutes for the time range of the last hour.
 After both hash set are reconciled, the Store protocol is used to reciprocally query the other node with the hash differences found.
 
--- a/standards/core/store-sync.md
+++ b/standards/core/store-sync.md
@@ -30,16 +30,17 @@ Various protocols and features that help with message consistency are described 
 This feature allow a node to fill the gap in messages for the period it was last offline.
 At startup, a node use the Store protocol to query a random node for
 the time interval since it was last online.
+Messages returned my the query are then added to the node local archive storage.
 It is RECOMMENDED” to limit the time interval to a maximum of 6 hours.
 
 ### Waku Sync
 
 Nodes that stay online can still miss messages.
-Waku Sync is used to find those messages by periodically syncing with random nodes.
-It is RECOMMENDED to trigger a sync every 5 minutes for the time range of the last hour.
-After both hash set are reconciled, the Store protocol is used to reciprocally query the other node with the hash differences found.
-
-TODO link to spec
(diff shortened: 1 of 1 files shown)


SionoiS:
Will require some changes when Waku sync 2.0 drops.

standards/core/store-sync.md
```diff
@@ -0,0 +1,52 @@
...
+The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”, “SHOULD”, “SHOULD NOT”, “RECOMMENDED”, “NOT RECOMMENDED”, “MAY”, and “OPTIONAL” in this document are to be interpreted as described in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt).
+
+Various protocols and features that help with message consistency are described below.
+
+### Store Resume
+
+This feature allow a node to fill the gap in messages for the period it was last offline.
+At startup, a node use the Store protocol to query a random node for
+the time interval since it was last online.
+Messages returned my the query are then added to the node local archive storage.
```
jm-clius, line 33:
```suggestion
Messages returned by the query are then added to the node local archive storage.
```

standards/core/store-sync.md
```diff
@@ -0,0 +1,52 @@
...
+It is RECOMMENDED” to limit the time interval to a maximum of 6 hours.
```
jm-clius, line 34:
```suggestion
It is RECOMMENDED to limit the time interval to a maximum of 6 hours.
```

standards/core/store-sync.md
```diff
@@ -0,0 +1,52 @@
...
+
+### Waku Sync
+
+Nodes that stay online can still miss messages.
+[Waku Sync](TODO link to spec) is the protocol used to find those messages by periodically syncing with random nodes.
```
jm-clius, line 39: Will be good to finish this TODO before merging.

standards/core/store-sync.md
```diff
@@ -0,0 +1,52 @@
...
+It is RECOMMENDED to trigger a sync every 5 minutes for a time range of the last hour.
+After both message hash sets are reconciled,
+the Store protocol is used to reciprocally query the other node with the message hash differences found.
+Messages returned my the query are then added to the node local Waku sync and archive storage.
```
jm-clius, line 43:
```suggestion
Messages returned by the query are then added to the node local Waku sync and archive storage.
```

standards/core/store-sync.md
jm-clius: Waku Sync, afaics will only be useful to find message _hashes_ and not messages, which is why Store queries are also required.

---

Perhaps not for this PR, but for following versions of the Waku Sync 2.0 protocol it's useful to think of a specification as a heuristic/set of directives that almost form a kind of pseudocode. This becomes easier if you use the RFC keywords for each directive, e.g:

Nodes participating in Store Sync MUST use Waku Sync to find diffs in message hashes with other Store Sync nodes.
The node MUST use Store queries to retrieve the full contents of messages for each message hash missing from its own set.

standards/core/store-sync.md
```diff
@@ -0,0 +1,52 @@
...
+At startup, a node use the Store protocol to query a random node for
+the time interval since it was last online.
+Messages returned my the query are then added to the node local archive storage.
+It is RECOMMENDED” to limit the time interval to a maximum of 6 hours.
+
+### Waku Sync
+
+Nodes that stay online can still miss messages.
+[Waku Sync](TODO link to spec) is the protocol used to find those messages by periodically syncing with random nodes.
+It is RECOMMENDED to trigger a sync every 5 minutes for a time range of the last hour.
```
jm-clius, line 40: I'd recommend separating the mechanism/principle from the implementation detail into two separate statements:
Nodes SHOULD trigger the store sync procedure regularly enough to maintain close to real-time consistency.
We RECOMMEND that sync is triggered every 5 minutes over the time range of the last hour.

jm-clius: What is the current RECOMMENDED number of nodes to sync with every 1 hour?

standards/core/store-sync.md
```diff
@@ -0,0 +1,52 @@
...
+
+The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”, “SHOULD”, “SHOULD NOT”, “RECOMMENDED”, “NOT RECOMMENDED”, “MAY”, and “OPTIONAL” in this document are to be interpreted as described in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt).
+
+Various protocols and features that help with message consistency are described below.
+
+### Store Resume
+
+This feature allow a node to fill the gap in messages for the period it was last offline.
+At startup, a node use the Store protocol to query a random node for
+the time interval since it was last online.
```
jm-clius, line 32: Perhaps only for future versions of the spec, but restating this as a heuristic rather than a high-level description of functionality allows the spec to be useful for new implementers, e.g.:

On startup, all Store Sync nodes SHOULD perform the resume procedure as set out below.
On startup the node MUST determine the last time it was online.
The node SHOULD do this by reading.... (?)
The Store Sync node MUST then craft a new store query to a random Store node
to retrieve all messages on relevant shards since the last time it was online.
The node MUST subsequently add these to the local storage.

jm-clius (approved):
Thanks! I've added some comments below. Since this is a raw spec that will be superseded very soon, I'm happy if we merge without all of the comments addressed.

SionoiS:
I added your comments to the document so that we don't lose them.

PR #39, opened 2024-09-13
chore: Adjusted peer-exchange to the latest changes made due to rate limit DOS protection

NagyZoltanPeter:
As a consequence of https://github.com/waku-org/nwaku/pull/3035 and https://github.com/waku-org/nwaku/issues/3028
Peer Exchange protocol specification must be changed.
In order to reflect possible error cases such as request is rejected due to rate limitation on responder side we must introduce `response_status` field which can hold detailed information of the processing outcome.

Status codes are also defined while `response_status` can contain descriptive information also.

### DIFF
--- a/standards/core/peer-exchange.md
+++ b/standards/core/peer-exchange.md
@@ -66,7 +66,7 @@ message PeerInfo {
   bytes enr = 1;
 }
 
-message PeerExchangeQuery {
+message PeerExchangeRequest {
   uint64 num_peers = 1;
 }
 
@@ -74,18 +74,37 @@ message PeerExchangeResponse {
   repeated PeerInfo peer_infos = 1;
 }
 
+message PeerExchangeResponseStatus {
+  uint32 status = 1;
+  optional string desc = 2;
+}
+
 message PeerExchangeRPC {
-  PeerExchangeQuery query = 1;
-  PeerExchangeResponse response = 2;
+  optional PeerExchangeRequest request = 1;
+  optional PeerExchangeResponse response = 2;
+  optional PeerExchangeResponseStatus status = 10;
 }
 
 ```
 
 The `enr` field contains a Waku ENR as specified in [WAKU2-ENR](./enr.md).
 
-Requesters send a `PeerExchangeQuery` to a peer.
+Requesters send a `PeerExchangeRequest` to a peer.
+Responders SHOULD not fill `request` field.
 Responders SHOULD include a maximum of `num_peers` `PeerInfo` instances into a response.
 Responders send a `PeerExchangeResponse` to requesters containing a list of `PeerInfo` instances, which in turn hold an ENR.
+Responders, MUST include a `PeerExchangeResponseStatus` in the response in any case. If the request was not successful `PeerExchangeResponse` can be omitted.
+
+### Possible status codes
+
+| Result | Code | Note |
+|--------|------|------|
+| SUCCESS  | 200 | Successfull request-respond. In response the answer must contain `PeerExchangeResponse`  |
+| BAD_REQUEST | 400   | Wrong request payload. It must only contain `reques` field.  |
+| BAD_RESPOND | 401   | Wrong respond payload. If success it must contain `respond` and `responseStatus` fields. If failure, only `responseStatus` is set.  |
+| TOO_MANY_REQUESTS | 429 | DOS protection prevented this request as the current request exceeds the configured request rate. |
+| SERVICE_UNAVAILABLE  | 503 | Request cannot be served, either issue on Responder side or having no suitable peer to issue request.  |
+| DIAL_FAILRE | 599 | Requester side problem calling PeerExchange peer. |
 
 ## Implementation Suggestions
 
@@ -143,6 +162,7 @@ Requesters send a simple message request which causes responders to engage in am
 As a mitigation, responders MAY feature a `seen cache` for requests and only answer once per time interval.
 The exchange-peer cache discussed in [Theory and Protocol Semantics Section](#theory-and-protocol-semantics) also provides mitigation.
 Still, frequent queries can tigger the refresh cycle more often. The `seen cache` MAY be used in conjunction to provide additional mitigation.
+Responders MAY apply request limits and thus can reject answering requests within a certain time window. Requestes must be prepared to handle this case.
 
 ### Further Considerations
 
--- a/standards/core/peer-exchange.md
+++ b/standards/core/peer-exchange.md
@@ -82,7 +82,7 @@ message PeerExchangeResponseStatus {
 message PeerExchangeRPC {
   optional PeerExchangeRequest request = 1;
   optional PeerExchangeResponse response = 2;
-  optional PeerExchangeResponseStatus status = 10;
+  optional PeerExchangeResponseStatus response_status = 10;
 }
 
 ```
@@ -93,7 +93,7 @@ Requesters send a `PeerExchangeRequest` to a peer.
 Responders SHOULD not fill `request` field.
 Responders SHOULD include a maximum of `num_peers` `PeerInfo` instances into a response.
 Responders send a `PeerExchangeResponse` to requesters containing a list of `PeerInfo` instances, which in turn hold an ENR.
-Responders, MUST include a `PeerExchangeResponseStatus` in the response in any case. If the request was not successful `PeerExchangeResponse` can be omitted.
+Responders, MUST include a `PeerExchangeResponseStatus` in the response in any case. If the request was not successful `response` field can be omitted.
 
 ### Possible status codes
 
@@ -101,7 +101,7 @@ Responders, MUST include a `PeerExchangeResponseStatus` in the response in any c
 |--------|------|------|
 | SUCCESS  | 200 | Successfull request-respond. In response the answer must contain `PeerExchangeResponse`  |
 | BAD_REQUEST | 400   | Wrong request payload. It must only contain `reques` field.  |
-| BAD_RESPOND | 401   | Wrong respond payload. If success it must contain `respond` and `responseStatus` fields. If failure, only `responseStatus` is set.  |
+| BAD_RESPOND | 401   | Wrong respond payload. If success it must contain `respond` and `response_status` fields. If failure, only `response_status` is set.  |
 | TOO_MANY_REQUESTS | 429 | DOS protection prevented this request as the current request exceeds the configured request rate. |
 | SERVICE_UNAVAILABLE  | 503 | Request cannot be served, either issue on Responder side or having no suitable peer to issue request.  |
 | DIAL_FAILRE | 599 | Requester side problem calling PeerExchange peer. |


standards/core/peer-exchange.md
```diff
@@ -66,26 +66,45 @@ message PeerInfo {
...
+message PeerExchangeRequest {
   uint64 num_peers = 1;
 }
 
 message PeerExchangeResponse {
   repeated PeerInfo peer_infos = 1;
 }
 
+message PeerExchangeResponseStatus {
+  uint32 status = 1;
```
chaitanyaprem, line 78: why not add these fields to existing PeerExchangeResponse itself? That way it makes it backward compatible.

standards/core/peer-exchange.md
```diff
@@ -66,26 +66,45 @@ message PeerInfo {
   bytes enr = 1;
 }
 
-message PeerExchangeQuery {
+message PeerExchangeRequest {
```
chaitanyaprem, line 69: Not sure if it breaks compatibility by renaming existing fields. Maybe better to keep same name?

chaitanyaprem (commented):
one more point to note is changes to this spec needs to be implemented in `go-waku` and hence need to go live together in status app and fleet nodes. Otherwise compatibility needs to be maintained so that even if we go live in fleet first, clients should not get affected. 

Also note that peer-exchange service is being provided by status-desktop as well which means update of fleet and desktop instances should be planned together if possible.

So, not sure if we want to take this live as part of status fleets or only as general protoocl update. just wanted to higlight this point.

standards/core/peer-exchange.md
NagyZoltanPeter: > why not add these fields to existing PeerExchangeResponse itself? That way it makes it backward compatible.

What do you mean? Isn't that the opposite? That would break the protocol. With this solution I tried to avoid breaking the prev protocol by adding it separately under a new index.

standards/core/peer-exchange.md
NagyZoltanPeter: It should not. I corrected this to follow what is used in nwaku originally. But for protobuf naming does not count but field indices, imho.

standards/core/peer-exchange.md
chaitanyaprem: afaiu adding new fields to existing messages doesn't break existing protocol [ref](https://stackoverflow.com/questions/45854645/google-protobuf-if-i-add-field-from-one-side-will-it-break-another-one), in fact that is the way protobufs are extendible in nature. New fields are ignored and not decoded by an older version.

standards/core/peer-exchange.md
chaitanyaprem: yea, looks that way...only if you change field numbers then it breaks the existing protocol.

standards/core/peer-exchange.md
NagyZoltanPeter: Ok, I think we are on the same page. Thank you for the research btw. I also had this in mind.
So I did not embed the status into the response to keep backward compatibility.
But as I went through the docs, which does not clearly indicate that extending an embedded type still maintains compatibility, now I read another sources whereas with protobuf3 all fields are like "optional" and will have a default (according to the type). So backward and forward compatibility is kept.

If that is acceptable I will change this as suggested (also in nwaku code) that rather have the status in the response field.
Also will remove the optional from the fields and leave to the callee/caller side to decide which they read.
Minimal checks can still be done through the status field.

Do you agree?

NagyZoltanPeter:
> > one more point to note is changes to this spec needs to be implemented in `go-waku` and hence need to go live together in status app and fleet nodes. Otherwise compatibility needs to be maintained so that even if we go live in fleet first, clients should not get affected.
> > Also note that peer-exchange service is being provided by status-desktop as well which means update of fleet and desktop instances should be planned together if possible.
> > So, not sure if we want to take this live as part of status fleets or only as general protoocl update. just wanted to higlight this point.
> 
> Yes, thank you for this point. I meant this PR also to talk about it openly. My idea was not to break existing functionality while able to add DOS protection to PeerEx. Also in nwaku implementation I found that there are error cases handled but no answer sent back at all to requester. This is fixed also in PR: [waku-org/nwaku#3035](https://github.com/waku-org/nwaku/pull/3035)
> 
> But in normal - previous - usage these change shall not affect clients using the old protocol. Feel free to review nwaku pr.

@chaitanyaprem: I applied the changes, please have a look.

standards/core/peer-exchange.md
```diff
@@ -66,26 +66,40 @@ message PeerInfo {
...
 
 ```
 
 The `enr` field contains a Waku ENR as specified in [WAKU2-ENR](./enr.md).
 
-Requesters send a `PeerExchangeQuery` to a peer.
+Requesters send a `PeerExchangeRequest` to a peer.
 Responders SHOULD include a maximum of `num_peers` `PeerInfo` instances into a response.
 Responders send a `PeerExchangeResponse` to requesters containing a list of `PeerInfo` instances, which in turn hold an ENR.
+Responders, MUST include a `status` in the response in any case and possible descriptive form of the problem if any.
```
chaitanyaprem, line 91: should this be change it to SHOULD so that clients can function with both new and older model where codes are not returned?

chaitanyaprem (approved):
LGTM, barring minor comment.

standards/core/peer-exchange.md
```diff
@@ -66,26 +66,40 @@ message PeerInfo {
...
+### Possible status codes
+
+| Result | Code | Note |
+|--------|------|------|
+| SUCCESS  | 200 | Successfull request-respond. In response the answer must contain `PeerExchangeResponse`  |
+| BAD_REQUEST | 400   | Wrong request payload. It must only contain `reques` field.  |
+| BAD_RESPOND | 401   | Wrong respond payload. If success it must contain `respond` and `response_status` fields. If failure, only `response_status` is set.  |
+| TOO_MANY_REQUESTS | 429 | DOS protection prevented this request as the current request exceeds the configured request rate. |
+| SERVICE_UNAVAILABLE  | 503 | Request cannot be served, either issue on Responder side or having no suitable peer to issue request.  |
+| DIAL_FAILRE | 599 | Requester side problem calling PeerExchange peer. |
```
jm-clius, line 102:
```suggestion
| DIAL_FAILURE | 599 | Requester side problem calling PeerExchange peer. |
```

standards/core/peer-exchange.md
```diff
@@ -66,26 +66,40 @@ message PeerInfo {
...
+Requesters send a `PeerExchangeRequest` to a peer.
 Responders SHOULD include a maximum of `num_peers` `PeerInfo` instances into a response.
 Responders send a `PeerExchangeResponse` to requesters containing a list of `PeerInfo` instances, which in turn hold an ENR.
+Responders, MUST include a `status` in the response in any case and possible descriptive form of the problem if any.
+
+### Possible status codes
+
+| Result | Code | Note |
+|--------|------|------|
+| SUCCESS  | 200 | Successfull request-respond. In response the answer must contain `PeerExchangeResponse`  |
```
jm-clius, line 97:
```suggestion
| SUCCESS  | 200 | Successful request-response. The response must contain a `PeerExchangeResponse`  |
```

Not sure I get the addition of the clause "The response must contain...". Since the status code that is enumerated here is _inside_ the `PeerExchangeResponse`, this clause is more apt as directive for the RPC? I'd just remove that clause

standards/core/peer-exchange.md
```diff
@@ -66,26 +66,40 @@ message PeerInfo {
...
+| BAD_REQUEST | 400   | Wrong request payload. It must only contain `reques` field.  |
```
jm-clius, line 98:
```suggestion
| BAD_REQUEST | 400   | Wrong request payload. It must only contain `request` field.  |
```
Again, here and elsewhere, not sure I understand the field directives here "It must only contain a `request`..." -> this is a field inside the `response`, so the existence of the response code violates its description unless I'm missing something. 😅

standards/core/peer-exchange.md
```diff
@@ -66,26 +66,40 @@ message PeerInfo {
...
+| BAD_RESPOND | 401   | Wrong respond payload. If success it must contain `respond` and `response_status` fields. If failure, only `response_status` is set.  |
```
jm-clius, line 99:
```suggestion
| BAD_RESPONSE | 401   | Wrong response payload. If success it must contain `response` and `response_status` fields. If failure, only `response_status` is set.  |
```

Unless this is for compatibility elsewhere, "respond" should be "response" in all field names. "Respond" is the verb, with "response" being the noun.

standards/core/peer-exchange.md
```diff
@@ -143,6 +157,7 @@ Requesters send a simple message request which causes responders to engage in am
 As a mitigation, responders MAY feature a `seen cache` for requests and only answer once per time interval.
 The exchange-peer cache discussed in [Theory and Protocol Semantics Section](#theory-and-protocol-semantics) also provides mitigation.
 Still, frequent queries can tigger the refresh cycle more often. The `seen cache` MAY be used in conjunction to provide additional mitigation.
+Responders MAY apply request limits and thus can reject answering requests within a certain time window. Requestes must be prepared to handle this case.
```
jm-clius, line 160:
```suggestion
Responders MAY apply request limits and thus can reject answering requests within a certain time window.
Requesters SHOULD handle this case.
```

standards/core/peer-exchange.md
```diff
@@ -66,26 +66,40 @@ message PeerInfo {
...
 
-message PeerExchangeQuery {
+message PeerExchangeRequest {
   uint64 num_peers = 1;
 }
 
 message PeerExchangeResponse {
   repeated PeerInfo peer_infos = 1;
+  uint32 status = 10;
+  optional string desc = 11;
```
jm-clius, line 76: Let's use the exact same field names as we have used for the new Lightpush and Filter protocols, namely:
`status_code` and `status_desc`.

jm-clius (changes requested):
Thanks! I have a couple of comments below, mostly minor and related to formulation. I think the most important to clarify is around field names. I think at some point the `status` and `desc` fields might have changed names from `respond` and `response_status` and this is not yet reflected in the description?

standards/core/peer-exchange.md
NagyZoltanPeter: Absolutely right, fixing it.

standards/core/peer-exchange.md
NagyZoltanPeter: Exactly, the previous version had status as separate field from response payload. But after discussing compatibility approaches with Prem, I changed the structure but this leftover remained here.

standards/core/peer-exchange.md
NagyZoltanPeter: Well, I'm not pretty sure. As - for me - a spec describes the state we are aiming and not the past we are coming from. So I would expect any implementation of the protocol to behave as "MUST" and not "SHOULD" or "MIGHT".
Of course for backward compatibility according to protobuf 3 specs these fields will get a default value according to their types... I believe this case `status` will take `0` and will mean UNKNOWN. So either old requester or responder will be able to operate. But sooner or later I expect them to follow the spec.
@chaitanyaprem , @jm-clius : WDYT?

standards/core/peer-exchange.md
NagyZoltanPeter: Ok.

standards/core/peer-exchange.md
NagyZoltanPeter: Removed that sentence.

standards/core/peer-exchange.md
```diff
@@ -66,26 +66,40 @@ message PeerInfo {
...
 
 ```
 
 The `enr` field contains a Waku ENR as specified in [WAKU2-ENR](./enr.md).
 
-Requesters send a `PeerExchangeQuery` to a peer.
+Requesters send a `PeerExchangeRequest` to a peer.
 Responders SHOULD include a maximum of `num_peers` `PeerInfo` instances into a response.
 Responders send a `PeerExchangeResponse` to requesters containing a list of `PeerInfo` instances, which in turn hold an ENR.
+Responders, MUST include a `status_code` in the response in any case and possible descriptive form of the problem if any.
```
jm-clius, line 91:
```suggestion
Responders, MUST include a `status_code` in the response.
They MAY include a descriptive status in the `status_desc` field if available.
```

standards/core/peer-exchange.md
```diff
@@ -66,26 +66,40 @@ message PeerInfo {
...
+
+### Possible status codes
+
+| Result | Code | Note |
+|--------|------|------|
+| SUCCESS  | 200 | Successfull request-respond.  |
```
jm-clius, line 97:
```suggestion
| SUCCESS  | 200 | Successful request-response.  |
```
English is weird - success + full = successful :)

standards/core/peer-exchange.md
jm-clius, line 160:
```suggestion
Responders MAY apply request limits and thus can reject answering requests within a certain time window.
Requesters must be prepared to handle this case.
```

jm-clius (approved):
Very minor comments, but can be merged when addressed

standards/core/peer-exchange.md
```diff
@@ -66,26 +66,40 @@ message PeerInfo {
...
+| BAD_REQUEST | 400   | Wrong request payload. It must only contain `request` field.  |
+| BAD_RESPONSE | 401   | Wrong response payload. If success it must contain `response` and `status_code` fields. If failure, only `status_code` is set.  |
```
chaitanyaprem, line 99: What does 401 mean? That server was not able to formulate a response. 
Also looks like description for this code is not that clear.

standards/core/peer-exchange.md
```diff
@@ -66,26 +66,40 @@ message PeerInfo {
...
+| TOO_MANY_REQUESTS | 429 | DOS protection prevented this request as the current request exceeds the configured request rate. |
+| SERVICE_UNAVAILABLE  | 503 | Request cannot be served, either issue on Responder side or having no suitable peer to issue request.  |
```
chaitanyaprem, line 101: `r having no suitable peer to issue request` is not valid for a peer-exchange server.
In case of peer-exchange the only scenario where 503 can be returned is when discv5 service is not running and no peers are populated in the cache of server.

standards/core/peer-exchange.md
```diff
@@ -66,26 +66,40 @@ message PeerInfo {
...
+| DIAL_FAILURE | 599 | Requester side problem calling PeerExchange peer. |
```
chaitanyaprem, line 102: is this a code that can be returned from server side? if not, may not be needed. 
i think this is something local that can happen at the client.

chaitanyaprem (approved):
few comments otherwise looks good.

standards/core/peer-exchange.md
NagyZoltanPeter: Yes, you're right, I make it clear

PR #45, opened 2024-10-28
Change first incentivized light protocol from Store to Lightpush

s-tikhomirov:
This PR changes the incentivization spec to outline Lightpush (instead of Store, as planned previously) to become the first incentivized light protocol in Waku.

### DIFF
--- a/standards/core/incentivization.md
+++ b/standards/core/incentivization.md
@@ -41,7 +41,7 @@ In that case, a client MUST provide a valid eligibility proof as part of its req
 Forms of eligibility proofs include:
 
 - Proof of payment: for paid non-authenticated requests. A proof of payment, in turn, may also take different forms, such as a transaction hash or a ZK-proof. In order to interpret a proof of payment, the server needs information about its type.
-- Proof of membership: for services for a predefined group of users. An example use case: an application developer pays in bulk for their users' requests. A client then prove that they belong to the user set of that application. Rate limiting in Waku RLN Relay is based a similar concept.
+- Proof of membership: for services for a predefined group of users. An example use case: an application developer pays in bulk for their users' requests. A client then prove that they belong to the user set of that application. Rate limiting in Waku RLN Relay is based on a similar concept.
 - Service credential: a proof of membership in a set of clients who have prepaid for the service (which may be considered a special case of proof of membership).
 
 Upon a receiving a request:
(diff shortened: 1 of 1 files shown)


standards/core/incentivization.md
```diff
@@ -41,13 +41,13 @@ In that case, a client MUST provide a valid eligibility proof as part of its req
 Forms of eligibility proofs include:
 
 - Proof of payment: for paid non-authenticated requests. A proof of payment, in turn, may also take different forms, such as a transaction hash or a ZK-proof. In order to interpret a proof of payment, the server needs information about its type.
-- Proof of membership: for services for a predefined group of users. An example use case: an application developer pays in bulk for their users' requests. A client then prove that they belong to the user set of that application. A similar concept (RLN) is used in Waku Relay for spam prevention.
+- Proof of membership: for services for a predefined group of users. An example use case: an application developer pays in bulk for their users' requests. A client then prove that they belong to the user set of that application. Rate limiting in Waku RLN Relay is based a similar concept.
```
jm-clius, line 44:
```suggestion
- Proof of membership: for services for a predefined group of users. An example use case: an application developer pays in bulk for their users' requests. A client then prove that they belong to the user set of that application. Rate limiting in Waku RLN Relay is based on a similar concept.
```

standards/core/incentivization.md
```diff
@@ -15,8 +15,8 @@ In an incentivized request-response protocol, only eligible (e.g., paying) clien
 Clients include eligibility proofs in their requests.
 
 Eligibility proofs are designed to be used in multiple Waku protocols, such as Store, Lightpush, and Filter.
-Store is planned to become the first Waku protocol to support incentivization.
-We discuss the proof-of-concept implementation of incentivization for Store in a later section.
+Lightpush is planned to become the first Waku protocol to support incentivization.
```
jm-clius, line 18: I think it's important to note that it's _attaching RLN proofs_ to lightpush requests that's the incentivised service here. It's perfectly possible for a client to generate and attach its own proofs and simply require the service to publish on its behalf, which is out of scope for now.

jm-clius (approved):
LGTM. Minor comments below.

standards/core/incentivization.md
s-tikhomirov: Good point, I'll clarify that.

PR #47, opened 2024-12-05
feat: waku sync 2.0

SionoiS:
WIP spec for Waku sync 2.0

### DIFF
--- a/standards/core/sync.md
+++ b/standards/core/sync.md
@@ -157,7 +157,7 @@ This section was created to inform implementations.
 Two useful parameters to add to your implementation are partitioning count and the item set threshold.
 
 The partitioning count is the number of time a range is split.
-Higher value reduce round trips at the cost of computing more fingerprints.
+A higher value reduces round trips at the cost of computing more fingerprints.
 
 The threshold for which item sets are sent instead of fingerprints.
 Higher value sends more items which means higher chance of duplicates but
--- a/standards/core/sync.md
+++ b/standards/core/sync.md
@@ -11,78 +11,131 @@ contributors:
 This specification explains the `WAKU-SYNC` protocol
 which enables the reconciliation of two sets of message hashes
 in the context of keeping multiple Store nodes synchronized.
-Waku Sync is a wrapper around
-[Negentropy](https://github.com/hoytech/negentropy) a [range-based set reconciliation protocol](https://logperiodic.com/rbsr.html).
 
 ## Specification
 
-**Protocol identifier**: `/vac/waku/sync/1.0.0`
+**Protocol identifier**: `/vac/waku/reconciliation/1.0.0`
 
-### Terminology
+#### Terminology
 The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”, “SHOULD”, “SHOULD NOT”, 
 “RECOMMENDED”, “MAY”, and “OPTIONAL” in this document are to be interpreted as described in [RFC2119](https://www.ietf.org/rfc/rfc2119.txt).
 
-The term Negentropy refers to the protocol of the same name.
-Negentropy payload refers to
-the messages created by the Negentropy protocol.
-Client always refers to the initiator
-and the server the receiver of the first payload.
+#### Message Ids
+Message Ids MUST be composed of the timestamp and the hash of the Waku messages.
 
-### Design Requirements
-Nodes enabling Waku Sync SHOULD
-manage and keep message hashes in a local cache
-for the range of time
-during which synchronization is required.
-Nodes SHOULD use the same time range,
-for Waku we chose one hour as the global default.
-Waku Relay or Light Push protocol MAY be enabled
-and used in conjunction with Sync
-as a source of new message hashes
-for the cache.
+The timestamp MUST be the time of creation and
+the hash MUST follow the
+[deterministic message hashing specification](https://rfc.vac.dev/waku/standards/core/14/message#deterministic-message-hashing)
 
-Nodes MAY use the Store protocol
-to request missing messages once reconciliation is complete
-or to provide messages to requesting clients.
+> This way the message Ids can always be totally ordered.
+Chronologically according to the timestamp and
+disambiguate based on the hash lexical order
+in cases where the timestamp is the same.
 
-### Payload
+#### Range Bounds
+A range MUST consists of 2 Id bounds, the first bound is
+inclusive the second bound exclusive.
+The first bound MUST be strictly smaller than the second one.
 
-```protobuf
-syntax = "proto3";
+#### Range Fingerprinting
+The fingerprint of a range MUST be the XOR operation applied to
+all the hashes of the Ids included in that range.
 
-package waku.sync.v1;
+#### Range Type
+Every range MUST have one of the following types; skip, fingerprint or item set.
 
-message SyncPayload {
-  optional bytes negentropy = 1;
+- Skip type is used to signal already processed ranges that MUST be ignored.
+- Fingerprint type signify that this range fingerprint MUST be compared when received.
+- Item set type contain multiple message Ids that MUST all be compared when received.
+> Item sets are an optimization, stopping the recursion early can
+save many network roundtrips.
 
-  repeated bytes hashes = 20;
-}
-```
+#### Range Processing
+Ranges have to be processed differently acording to their types.
 
-### Session Flow
-A client initiates a session with a server
-by sending a `SyncPayload` with
-only the `negentropy` field set.
-This field MUST contain
-the first negentropy payload
-created by the client
-for this session.
+- Skip ranges MUST be merged with other consequtive ones if possible.
+- Equal fingerprint ranges MUST become skip ranges.
+- Unequal fingerprint ranges MUST be splitted into smaller ranges. The new type MAY be either fingerprint or item set.
+- Unresolved item set ranges MUST be checked for differences and marked resolved.
+- Resolved item set ranges MUST be checked for differences and become skip ranges.
 
-The server receives a `SyncPayload`.
-A new negentropy payload is computed from the received one.
-The server sends back a `SyncPayload` to the client.
+### Delta Encoding
+For efficient transmition of timestamps, hashes and ranges. Payloads are delta encoded as follow.
 
-The client receives a `SyncPayload`.
-A new negentropy payload OR an empty one is computed.
-If a new payload is computed then
-the exchanges between client and server continues until
-the client computes an empty payload.
-This client computation also outputs any hash differences found,
-those MUST be stored.
-In the case of an empty payload,
-the reconciliation is done,
-the client MUST send back a `SyncPayload`
-with all the missing server hashes in the `hashes` field and
-an empty `nengentropy` field.
+All ranges to be transmitted MUST be ordered and only upper bounds used.
+> Inclusive lower bounds can be omitted because they are always
+the same as the exclusive upper bounds of the previous range or zero.
+
+To achieve this, it MAY be needed to add skip ranges.
+> For example, a skip range can be added with
+an exclusive upper bound equal to the first range lower bound.
+This way the receiving peer knows to ignore the range from zero to the start of the ranges
+
+Every timestamps after the first MUST be noted as the difference from the previous one.
+If the timestamp is the same, zero MUST be used and the hash MUST be added.
+The added hash MUST be trucated up to and including the first differetiating byte.
+
+| Timestamp | Hash | Timestamp (encoded) | Hash (encoded) 
+| - | - | - | -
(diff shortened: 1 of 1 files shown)


SionoiS:
I still don't know what to write for varints.

I don't want to write a varint lib and I also think the Nim std impl. might change.

What do? @jm-clius

standards/core/sync.md
```diff
@@ -7,126 +7,186 @@ contributors:
...
+
+#### Message Ids
+Message Ids MUST be composed of the timestamp and the hash of the Waku messages.
+
+The timestamp MUST be the time of creation and
+the hash MUST follow the
+[deterministic message hashing specification](https://rfc.vac.dev/waku/standards/core/14/message#deterministic-message-hashing)
+
+> This way the message Ids can always be totally ordered.
+Chronologically according to the timestamp and
```
jm-clius, line 37: The message IDs are ordered chronologically,....

(Sentence needs a subject and verb)

standards/core/sync.md
```diff
@@ -7,126 +7,186 @@ contributors:
...
+A range MUST consists of 2 Id bounds, the first bound is
+inclusive the second bound exclusive.
+The first bound MUST be strictly smaller than the second one.
+
+#### Range Fingerprinting
+The fingerprint of a range MUST be the XOR operation applied to
+all the hashes of the messages included in that range.
+
+#### Range Type
+Every range MUST have one of the following types; skip, fingerprint or item set.
```
jm-clius, line 51: Nitpick, but could to typographically highlight repeated concepts

```suggestion
Every range MUST have one of the following types; `skip`, `fingerprint` or `item set`.
```

standards/core/sync.md
```diff
@@ -7,126 +7,186 @@ contributors:
...
+Ranges have to be processed differently according to their types.
+
+- Skip ranges MUST be merged with other consecutive ones if possible.
+- Equal fingerprint ranges MUST become skip ranges.
+- Unequal fingerprint ranges MUST be splitted into smaller ranges. The new type MAY be either fingerprint or item set.
+- Unresolved item set ranges MUST be checked for differences and marked resolved.
+- Resolved item set ranges MUST be checked for differences and become skip ranges.
+
+### Delta Encoding
+For efficient transmission of timestamps, hashes and ranges. Payloads are delta encoded as follow.
```
jm-clius, line 69:
```suggestion
Payloads MUST be delta encoded as follows.
for efficient transmission of timestamps, hashes and ranges. 
```

Nitpick here.

standards/core/sync.md
```diff
@@ -7,126 +7,186 @@ contributors:
...
-When differences are found some messages will have to be inserted randomly.
-It is expected to be a less likely case than time based insertion and removal.
-Last but not least it must be optimized for sequential read
-as it is the most often used operation.
+#### Peer Choice
+Wrong peering strategies can lead to inadvertently segregating peers and
+reduce sampling diversity.
+Nwaku randomly select peers to sync with for simplicity and robustness.
+
+Good strategies can be devised but we chose not to.
```
jm-clius, line 174: I would simply say something like:

```suggestion
More sophisticated strategies may be implemented in future.
```

jm-clius (changes requested):
Thanks for this!

I left some comments below, many of them nitpicks.
I think the main comment is around providing some context and a functional overview before jumping into the detail (at least of the `reconciliation` protocol). It can state the protocol flow and wire protocols in very simple terms:
1. what agents are involved (two nodes that store Waku messages)
2. what are the preconditions and goal (they want to initiate a bidirectional sync over the messages)
3. what SHOULD and MUST they do, step by step, to achieve this:
- they identify the _ranges_ that they want to sync over (and define what is meant by that)
- they create the fingerprint
- they encode in some way

etc. 

I think this can be done in a very short paragraph, but will help first-time readers understand the big picture.

standards/core/sync.md
```diff
@@ -7,126 +7,186 @@ contributors:
...
+# Abstract
+This specification explains `WAKU-SYNC`
+which enables the synchronization of messages between 2 Store nodes.
 
-## Specification
+# Specification
 
-**Protocol identifier**: `/vac/waku/sync/1.0.0`
+Waku Sync consists of 2 protocols; reconciliation and transfer.
+Reconciliation is the process of finding differences in 2 sets of message hashes.
```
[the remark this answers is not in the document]
SionoiS: I'll try to clarify. Indeed fingerprints are of hashes only and sync ranges are IDs based.

standards/core/sync.md
```diff
@@ -7,126 +7,204 @@ contributors:
...
+- _Skip_ type contains nothing and is used to signal already processed _ranges_.
+- _Item set_ type contains message _Ids_ and a _resolved_ boolean.
+> _Item sets_ are an optimization, sending multiple _IDs_ instead of
+recursing further reduce the number of round-trips.
+
+#### Range Processing
+_Ranges_ have to be processed differently according to their types.
+
+- **Equal** _fingerprint ranges_ MUST become _skip ranges_.
+- **Unequal** _fingerprint ranges_ MUST be splitted into smaller ranges. The new type MAY be either _fingerprint_ or _item set_.
```
jm-clius, line 79:
```suggestion
- **Unequal** _fingerprint ranges_ MUST be split into smaller ranges. The new type MAY be either _fingerprint_ or _item set_.
```

jm-clius: Shall we add some guidance here as to when the new type is another fingerprint and when it is an item set? We don't need numbers, unless it's hardcoded, but we could just specify that when the new range is "small enough" the individual IDs can be encoded into an item set.

standards/core/sync.md
```diff
@@ -7,126 +7,204 @@ contributors:
...
+- **Unresolved** _item set_ ranges MUST be checked for differences and marked resolved.
```
jm-clius, line 80: What is the output of "checked for differences"? Does this peer just locally find and keep a list of missing IDs to be requested via transfer or does it also find the list of missing IDs in remote peer and send that in a subsequent message? I think probably the latter, but good idea to be explicit here: Unresolved item set...checked for differences. The IDs missing locally are ... (kept to be requested via transfer)?; the IDs missing in the remote peer are...(compiled and sent in a resolved set?); the item set is marked resolved (and sent to the remote peer?)

standards/core/sync.md
```diff
@@ -7,126 +7,204 @@ contributors:
...
+- **Resolved** _item set_ ranges MUST be checked for differences and become skip ranges.
```
jm-clius, line 81: Is it possible to receive a resolved set simply because this set has been resolved in the remote peer? In other words, it's not resolved by _this_ peer yet?

standards/core/sync.md
```diff
@@ -7,126 +7,204 @@ contributors:
...
+#### Payload encoding
+The wire level payload MUST be encoded as follow.
+> The & denote concatenation.
+
+1. _varint_ bytes of the delta encoded timestamp &
+2. if timestamp is zero, 1 byte for the hash bytes length & the hash bytes &
+3. 1 byte, the _range_ type &
+4. either
+    - 32 bytes _fingerprint_ &
+    - _varint_ bytes of the item set length & bytes of every items &
```
jm-clius, line 121: Question, rather than a suggestion - the `&` makes it seem like these items will be concatenated, but if I understand correctly only one of the three listed are added. Perhaps "or" is a bit clearer after "either"?

```suggestion
    - 32 bytes _fingerprint_ or
    - _varint_ bytes of the item set length & bytes of every items or
```

standards/core/sync.md
```diff
@@ -7,126 +7,204 @@ contributors:
...
 
-Further refinements to the protocol are planned
-to reduce the trust level required to operate.
-Notably by verifying messages RLN proof at reception.
+#### Parameters 
+Two useful parameters to add to your implementation are partitioning count and the item set threshold.
 
-## Implementation
-The following is not part of the specifications but good to know implementation details.
+The partitioning count is the number of time a range is splitted.
```
jm-clius, line 159:
```suggestion
The partitioning count is the number of time a range is split.
```

standards/core/sync.md
```diff
@@ -7,126 +7,204 @@ contributors:
...
+Higher value reduce round trips at the cost of computing more fingerprints.
```
jm-clius, line 160:
```suggestion
A higher value reduces round trips at the cost of computing more fingerprints.
```

standards/core/sync.md
jm-clius: Update: you can even point to the implementation section on "item set threshold". Here you can just mention that if number of items in a range fall below _a_ threshold, it becomes an item set rather than just a fingerprint.

standards/core/sync.md
```diff
@@ -7,126 +7,204 @@ contributors:
...
 
-### Peer Choice
-Peering strategies can lead to inadvertently segregating peers and reduce sampling diversity.
-We randomly select peers to sync with for simplicity and robustness.
+The threshold for which item sets are sent instead of fingerprints.
```
jm-clius, line 162:
```suggestion
The item set threshold determines when item sets are sent instead of fingerprints.
```

standards/core/sync.md
```diff
@@ -7,126 +7,204 @@ contributors:
...
+Higher value sends more items which means higher chance of duplicates but
```
jm-clius, line 163:
```suggestion
A higher value sends more items which means higher chance of duplicates but
```

standards/core/sync.md
```diff
@@ -7,126 +7,204 @@ contributors:
...
+reduce the amount of round trips overall.
```
jm-clius, line 164:
```suggestion
reduces the amount of round trips overall.
```

jm-clius (approved):
I think this reads very well. Thanks! I have added some comments below, most of them nitpicks or phrasing suggestions which are up to you to address or not.

standards/core/sync.md
```diff
@@ -7,126 +7,204 @@ contributors:
...
-[Negentropy](https://github.com/hoytech/negentropy) a [range-based set reconciliation protocol](https://logperiodic.com/rbsr.html).
+# Abstract
+This specification explains `WAKU-SYNC`
+which enables the synchronization of messages between nodes storing sets of [`14/WAKU2-MESSAGE`](https://rfc.vac.dev/waku/standards/core/14/message)
 
-## Specification
+# Specification
 
-**Protocol identifier**: `/vac/waku/sync/1.0.0`
+Waku Sync consists of 2 libp2p protocols; reconciliation and transfer.
```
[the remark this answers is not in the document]
SionoiS: Good advice. Thanks!

standards/core/sync.md
```diff
@@ -7,126 +7,204 @@ contributors:
...
-or to provide messages to requesting clients.
-
-### Payload
+## Reconciliation
 
+**Libp2p Protocol identifier**: `/vac/waku/reconciliation/1.0.0`
+
+The protocol finds differences between 2 peers by
+comparing _fingerprints_ of _ranges_ of message _IDs_.
+_Ranges_ are encoded into payloads, exchanged between the peers and when the range _fingerprints_ are different, splitted into smaller ones.
```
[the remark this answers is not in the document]
SionoiS: Looks like `splitted` is not a word. The more you know...

standards/core/sync.md
```diff
@@ -7,126 +7,204 @@ contributors:
...
+at this point _IDs_ are sent for comparison instead of _fingerprints_.
+
+#### Overview
+
+1. The requestor choose a time range to sync.
+2. The range is encoded into a payload and sent.
+3. The requestee receive the payload and decode it.
+4. The range is processed and more ranges produced.
+5. The new ranges are encoded and sent.
+6. Payloads are repeatedly exchanged and differences between the peers are discovered.
```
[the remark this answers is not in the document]
SionoiS: `until` is misleading here. I'll try to find a new formulation.

standards/core/sync.md
```diff
@@ -7,126 +7,204 @@ contributors:
...
+7. The synchronization is done when no ranges are left to process.
```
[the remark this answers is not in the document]
SionoiS: Well the process is the same. Diffs may be found but nothing changes. I'll try to rephrase.

standards/core/sync.md
SionoiS: I'm not sure I understand.

An unequal fingerprint range can be received and produce an unresolved item set range.
An unresolved item set range can be received and produce a resolved item set range.
A resolved item set can be received and produces a skip range.

Those are the possible cases.

SionoiS: Input -> process -> output

You don't process your own output.

standards/core/sync.md
SionoiS: Both actually, all differences are sent to the transfer protocol.

I'll try to make this clearer.

PR #50, opened 2025-01-30
chore: waku sync cluster & shards

SionoiS:


jm-clius (approved):
Ah! Missed it previously. Thanks! :)

PR #53, opened 2025-02-13
waku mix integration spec

chaitanyaprem:
Document explaining mix integration with waku.

This is still work in progress.

@jm-clius could you take a look at structure and suggest anything else that would be required to be added?

### DIFF
--- a/standards/core/mix.md
+++ b/standards/core/mix.md
@@ -1,0 +1,128 @@
+---
+title: WAKU-MIX
+name: Waku Mix
+editor: Prem Chaitanya Prathi <premprathi@proton.me>
+contributors:
+ - Akshaya Mani <akshaya@status.im>
+ - Hanno Cornelius <hanno@status.im>
+---
+
+## Tags
+
+`waku/core-protocol`
+
+# Abstract
+The document describes [libp2p mix](https://rfc.vac.dev/vac/raw/mix/) integration into waku.
+This integration provides higher anonymity for users publishing or querying for messages to/from the Waku network.
+
+This document covers integration of mix with [lightpush](https://rfc.vac.dev/waku/standards/core/19/lightpush) and [store](https://rfc.vac.dev/waku/standards/core/13/store) protocols.
+Both of these are `request-response` based protocols that follow a service-user and a service-provider model.
+A node that initiates the request is a `service-user/client` whereas a node that replies to the request is the `service-provider/service-node`.
+
+This document also covers the aspect of relay nodes acting as mix nodes.
+Integration of [waku relay](https://rfc.vac.dev/waku/standards/core/11/relay) with mix is not covered in this document.
+
+## Background / Rationale / Motivation
+
+Waku protocols have weak sender/originator anonymity as explained in [Waku Privacy and Anonymity Analysis](https://vac.dev/rlog/wakuv2-relay-anon/).
+Without further anonymization, it is easy to analyze the network traffic and determine the originator of messages published into the network.
+Topic interests of a user can be identified by analyzing [Store](https://rfc.vac.dev/waku/standards/core/13/store) query messages sent by a user.
+Same applies for topic subscriptions via [Filter](https://rfc.vac.dev/waku/standards/core/12/filter) protocol.
+
+`Mix` protocol allows libp2p nodes to send messages without revealing the sender's identity (peer ID, IP address) to intermediary mix nodes or the recipient/destination.
+Anonymity is achieved by using the [Sphinx packet format](#references), which encrypts and routes messages through a series of mix nodes before reaching the recipient.
+
+By integrating the `mix` protocol into the waku network, we can improve the anonymity for publishers and store query users.
+Each waku relay node will be acting as a mix node that forms an `overlay mix network`.
+This network of mix nodes SHALL relay mix messages anonymously to the recepient.
+
+Anonymity of [Filter](https://rfc.vac.dev/waku/standards/core/12/filter) users is not addressed by this document.
+
+## Terminology
+The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”, “SHOULD”, “SHOULD NOT”,
+“RECOMMENDED”, “MAY”, and “OPTIONAL” in this document are to be interpreted as described in [RFC2119](https://www.ietf.org/rfc/rfc2119.txt).
+
+## Theory / Semantics
+
+Waku Mix creates an `overlay network` of all the waku nodes that support the `mix` protocol.
+
+Nodes with `mix` protocol mounted SHOULD advertise that they support mix protocol via their their chosen discovery method.
+They MAY do so by updating their [ENR](#enr-updates) and using one of the ENR based discovery methods.
+
+Nodes that want higher anonymity while publishing a message via [lightpush](https://rfc.vac.dev/waku/standards/core/19/lightpush) or performing a [store](https://rfc.vac.dev/waku/standards/core/13/store) query SHOULD use the `mix` protocol to route their messages to the destination.
+Sender Node that would like to use `mix` protocol SHOULD discover enough mix nodes so that there is always a healthy pool of mix nodes available for selection.
+The pool size of mix nodes SHOULD be large enough for the mixing to be effective.
+We RECOMMEND a pool size of at least 100 mix nodes for the mixing to be effective.
+
+The serialized [Waku Message](https://rfc.vac.dev/waku/standards/core/14/message) SHOULD be the payload in the [sphinx packet](https://rfc.vac.dev/vac/raw/mix#4-sphinx-packet-format).
+
+In order to provide anonymity of the sender node from the `service-node`,  `Single Use reply blocks` or `anonymous replies` as specified in the original [sphinx](#references) paper SHALL be used.
+
+A node that sends messages using `mix` MAY use two redundant (paths)[https://rfc.vac.dev/vac/raw/mix#24-node-discovery] to have better reliability of the message being delivered.
+It is up to the higher-layer mixed protocol to deduplicate redundant messages received in this way.
+
+## Node Roles
+
+Mix protocol defines 3 roles for the nodes in the mix network - `sender`, `exit`, `intermediary`.
+
+- An `sender` node is the originator node of a message, i.e a node that wishes to publish/query messages to/from the waku network.
+- An `exit` node is responsible for delivering messages to destination peer in the network.
+- An `intermediary` node is responsible for forwarding a mix packet to the next mix node in the path.
+
+A Waku relay node SHOULD by default have mix `intermediary` and `exit` node roles in the network.
+The implementation MAY provide a configuration to disable a node from acting as an `intermediary\exit` node.
+
+Any waku node that wishes to publish/query messages from the waku network MUST act as a `sender` node.
+
+Resource-restricted/Edge nodes with short connection windows SHOULD always be acting as `sender` nodes.
+
+## ENR updates
+
+Each waku node that supports `mix` SHOULD indicate the same in its discoverable [ENR](https://github.com/waku-org/specs/blob/master/standards/core/enr.md).
+The following fields MUST be set as part of the discoverable ENR of a mix waku node:
+- The `bit 5` in the [waku2 ENR key](https://github.com/waku-org/specs/blob/master/standards/core/enr.md#waku2-enr-key) is reserved to indicate `mix`support. This bit MUST be set to true to indicate `mix` support.
+- A new field `mix-key` SHOULD be set to the `ed25519 public key` which is used for sphinx encryption.
+
+### Discovery
+
+Mix protocol provides better anonymity when an originator node has a sufficiently large pool of mix nodes to do path selection.
+This moves the problem into discovery domain and requires the following from discovery mechanisms:
+1. It is important for nodes to be able to discover as many nodes as possible quickly. This becomes especially important for edge nodes that come online just to publish/query messages for a short period of time.
+2. It is important to have the most recent online status of the nodes so that mix paths that are selected are not broken which lead to reliability issues.
+
+Point-2 above can be mitigated partially by choosing redundant mix paths for the same message by the sender node.
+This may not be an effective solution as it increases the overall bandwidth usage.
+
+## Spam Protection
+
+Mix protocol in waku network SHOULD have `rate-limiting/spam` protection to handle scenarios such as below:
+
+1. Any node can generate a mix packet and publish into the mix network. Hence there needs to be some validation as to who is allowed to publish and whether the user is within allowed rate-limits.
+2. Any node can intentionally generate paths which are broken and send messages into the mix network.
+3. An attacker can spawn a huge number of mix nodes so that user behaviour is observed in order to determine traffic patterns and deanonymize users. 
+
+There is a need to enforce rate-limits and spam protect the mix network. 
+The rate-limiting and spam protection SHALL be addressed as part of future work.
+
+## Tradeoffs
+
+Using `mix` protocol for publishing and querying messages adds certain overhead which is primarily the delay in delivering message to the destination.
+The overall additional delay `D` depends on the following params:
+- path length `L`
+- delay added by each intermediary node `dm`
+- connection establishment time `dc`
+- processing delay `dp`
+
+Delay overhead can be calculated as `D =  L * (dm + dc +dp)`
+
+## Copyright
+
+Copyright and related rights waived via
+[CC0](https://creativecommons.org/publicdomain/zero/1.0/).
+
+## References
+ - [libp2p mix](https://rfc.vac.dev/vac/raw/mix/)
+ - [waku lightpush](https://rfc.vac.dev/waku/standards/core/19/lightpush)
+ - [waku relay](https://rfc.vac.dev/waku/standards/core/11/relay)
+ - [ENR](https://github.com/waku-org/specs/blob/master/standards/core/enr.md)
+ - [sphinx encryption](https://cypherpunks.ca/~iang/pubs/Sphinx_Oakland09.pdf)
--- a/standards/core/mix.md
+++ b/standards/core/mix.md
@@ -54,7 +54,7 @@ Sender Node that would like to use `mix` protocol SHOULD discover enough mix nod
 The pool size of mix nodes SHOULD be large enough for the mixing to be effective.
 We RECOMMEND a pool size of at least 100 mix nodes for the mixing to be effective.
 
-The serialized [Waku Message](https://rfc.vac.dev/waku/standards/core/14/message) SHOULD be the payload in the [sphinx packet](https://rfc.vac.dev/vac/raw/mix#4-sphinx-packet-format).
+The serialized [Waku Message](https://rfc.vac.dev/waku/standards/core/14/message) MUST be the payload in the [sphinx packet](https://rfc.vac.dev/vac/raw/mix#4-sphinx-packet-format).
 
 In order to provide anonymity of the sender node from the `service-node`,  `Single Use reply blocks` or `anonymous replies` as specified in the original [sphinx](#references) paper SHALL be used.
 
@@ -78,9 +78,9 @@ Resource-restricted/Edge nodes with short connection windows SHOULD always be ac
 
 ## ENR updates
 
-Each waku node that supports `mix` SHOULD indicate the same in its discoverable [ENR](https://github.com/waku-org/specs/blob/master/standards/core/enr.md).
+Each waku node that supports the `mix intermediary or exit role` SHOULD indicate the same in its discoverable [ENR](https://github.com/waku-org/specs/blob/master/standards/core/enr.md).
 The following fields MUST be set as part of the discoverable ENR of a mix waku node:
-- The `bit 5` in the [waku2 ENR key](https://github.com/waku-org/specs/blob/master/standards/core/enr.md#waku2-enr-key) is reserved to indicate `mix`support. This bit MUST be set to true to indicate `mix` support.
+- The `bit 5` in the [waku2 ENR key](https://github.com/waku-org/specs/blob/master/standards/core/enr.md#waku2-enr-key) is reserved to indicate `mix` support. This bit MUST be set to true to indicate `mix` support.
 - A new field `mix-key` SHOULD be set to the `ed25519 public key` which is used for sphinx encryption.
 
 ### Discovery
@@ -102,7 +102,7 @@ Mix protocol in waku network SHOULD have `rate-limiting/spam` protection to hand
 3. An attacker can spawn a huge number of mix nodes so that user behaviour is observed in order to determine traffic patterns and deanonymize users. 
 
 There is a need to enforce rate-limits and spam protect the mix network. 
-The rate-limiting and spam protection SHALL be addressed as part of future work.
+The rate-limiting and spam protection shall be addressed as part of future work.
 
 ## Tradeoffs
 
--- a/standards/core/mix.md
+++ b/standards/core/mix.md
@@ -33,7 +33,7 @@ Same applies for topic subscriptions via [Filter](https://rfc.vac.dev/waku/stand
 Anonymity is achieved by using the [Sphinx packet format](#references), which encrypts and routes messages through a series of mix nodes before reaching the recipient.
 
 By integrating the `mix` protocol into the waku network, we can improve the anonymity for publishers and store query users.
-Each waku relay node will be acting as a mix node that forms an `overlay mix network`.
+Each waku relay node SHOULD be acting as a mix node that forms an `overlay mix network`.
 This network of mix nodes SHALL relay mix messages anonymously to the recepient.
 
 Anonymity of [Filter](https://rfc.vac.dev/waku/standards/core/12/filter) users is not addressed by this document.
@@ -44,9 +44,9 @@ The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL 
 
 ## Theory / Semantics
 
-Waku Mix creates an `overlay network` of all the waku nodes that support the `mix` protocol.
+Waku Mix creates an `overlay network` of all the Waku nodes that support the `mix` protocol.
 
-Nodes with `mix` protocol mounted SHOULD advertise that they support mix protocol via their their chosen discovery method.
+Nodes with `mix` protocol mounted SHOULD advertise that they support `mix` protocol via their their chosen discovery method.
 They MAY do so by updating their [ENR](#enr-updates) and using one of the ENR based discovery methods.
 
 Nodes that want higher anonymity while publishing a message via [lightpush](https://rfc.vac.dev/waku/standards/core/19/lightpush) or performing a [store](https://rfc.vac.dev/waku/standards/core/13/store) query SHOULD use the `mix` protocol to route their messages to the destination.
@@ -72,7 +72,7 @@ Mix protocol defines 3 roles for the nodes in the mix network - `sender`, `exit`
 A Waku relay node SHOULD by default have mix `intermediary` and `exit` node roles in the network.
 The implementation MAY provide a configuration to disable a node from acting as an `intermediary\exit` node.
 
-Any waku node that wishes to publish/query messages from the waku network MUST act as a `sender` node.
+Any waku node that wishes to publish/query messages via `mix` from the waku network MUST act as a `sender` node.
 
 Resource-restricted/Edge nodes with short connection windows SHOULD always be acting as `sender` nodes.
 
@@ -85,7 +85,7 @@ The following fields MUST be set as part of the discoverable ENR of a mix waku n
 
 ### Discovery
 
-Mix protocol provides better anonymity when an originator node has a sufficiently large pool of mix nodes to do path selection.
+Mix protocol provides better anonymity when a sender node has a sufficiently large pool of mix nodes to do path selection.
 This moves the problem into discovery domain and requires the following from discovery mechanisms:
 1. It is important for nodes to be able to discover as many nodes as possible quickly. This becomes especially important for edge nodes that come online just to publish/query messages for a short period of time.
 2. It is important to have the most recent online status of the nodes so that mix paths that are selected are not broken which lead to reliability issues.
--- a/standards/core/mix.md
+++ b/standards/core/mix.md
@@ -20,7 +20,6 @@ Both of these are `request-response` based protocols that follow a service-user 
 A node that initiates the request is a `service-user/client` whereas a node that replies to the request is the `service-provider/service-node`.
 
 This document also covers the aspect of relay nodes acting as mix nodes.
-Integration of [waku relay](https://rfc.vac.dev/waku/standards/core/11/relay) with mix is not covered in this document.
 
 ## Background / Rationale / Motivation
 
@@ -83,14 +82,17 @@ The following fields MUST be set as part of the discoverable ENR of a mix waku n
 - The `bit 5` in the [waku2 ENR key](https://github.com/waku-org/specs/blob/master/standards/core/enr.md#waku2-enr-key) is reserved to indicate `mix` support. This bit MUST be set to true to indicate `mix` support.
 - A new field `mix-key` SHOULD be set to the `ed25519 public key` which is used for sphinx encryption.
 
+Adding these fields to the ENR may cause the ENR size to cross the 300 byte limit especially in case a node supports multiple transports. 
+This limitation will have to be addressed in future.
 ### Discovery
 
 Mix protocol provides better anonymity when a sender node has a sufficiently large pool of mix nodes to do path selection.
 This moves the problem into discovery domain and requires the following from discovery mechanisms:
 1. It is important for nodes to be able to discover as many nodes as possible quickly. This becomes especially important for edge nodes that come online just to publish/query messages for a short period of time.
-2. It is important to have the most recent online status of the nodes so that mix paths that are selected are not broken which lead to reliability issues.
+2. The discovery mechanism MUST be unbiased and not biased toward specific subsets (e.g., nodes that are topologically closer).
+3. It is important to have the most recent online status of the nodes so that mix paths that are selected are not broken which lead to reliability issues.
 
-Point-2 above can be mitigated partially by choosing redundant mix paths for the same message by the sender node.
+Point-3 above can be mitigated partially by choosing redundant mix paths for the same message by the sender node.
 This may not be an effective solution as it increases the overall bandwidth usage.
 
 ## Spam Protection
@@ -114,6 +116,12 @@ The overall additional delay `D` depends on the following params:
 - processing delay `dp`
 
 Delay overhead can be calculated as `D =  L * (dm + dc +dp)`
+
+## Future Work
+
+- Integration of [waku relay](https://rfc.vac.dev/waku/standards/core/11/relay) with mix.
+- [Spam protection](#spam-protection) of mix overlay network.
+- Alternative way to communicate mix-public-key in order to overcome [ENR](#enr-updates) size limitations.
 
 ## Copyright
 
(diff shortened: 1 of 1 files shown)


standards/core/mix.md
```diff
@@ -0,0 +1,119 @@
...
+title: WAKU-MIX
+name: Waku Mix
+editor: Prem Chaitanya Prathi <premprathi@proton.me>
+contributors:
+ - Akshaya Mani <akshaya@status.im>
+ - Hanno Cornelius <hanno@status.im>
+---
+
+# Abstract
+This document explains how [libp2p mix](https://rfc.vac.dev/vac/raw/mix/) protocol will be integrated into waku. This would provide higher anonymity for users publishing or querying for messages to/from the Waku network.
```
jm-clius, line 11: Minor note: specifications generally don't describe an intent ("will be integrated"), but specifies a newly designed reality (even if implementation has to catch up). In other words, "the document describes mix integration into Waku".

standards/core/mix.md
```diff
@@ -0,0 +1,119 @@
...
+editor: Prem Chaitanya Prathi <premprathi@proton.me>
+contributors:
+ - Akshaya Mani <akshaya@status.im>
+ - Hanno Cornelius <hanno@status.im>
+---
+
+# Abstract
+This document explains how [libp2p mix](https://rfc.vac.dev/vac/raw/mix/) protocol will be integrated into waku. This would provide higher anonymity for users publishing or querying for messages to/from the Waku network. 
+
+This document covers integration of mix with [lightpush](https://rfc.vac.dev/waku/standards/core/19/lightpush) and [store](https://rfc.vac.dev/waku/standards/core/13/store) protocols. It also covers the aspect of relay nodes acting as mix nodes.
```
jm-clius, line 13:
```suggestion
This document covers integration of mix with [lightpush](https://rfc.vac.dev/waku/standards/core/19/lightpush) and [store](https://rfc.vac.dev/waku/standards/core/13/store) protocols.
It also covers the aspect of relay nodes acting as mix nodes.
```

Please use semantic breaks (https://sembr.org/) throughout.

standards/core/mix.md
```diff
@@ -0,0 +1,119 @@
...
+Both of these are request response based protocols that follow a service-user and service-provider model.
+
+Integration of gossipsub with mix is not covered in this document.
+
+## Background / Rationale / Motivation
+
+Waku protocols have weak sender/originator anonymity as explained in [Waku Privacy and Anonymity Analysis](https://vac.dev/rlog/wakuv2-relay-anon/).
+As of today, it is easy to analyze the network traffic and determine originator of messages published into the network or determine the topic interests by analyzing store queries requested by a user. Similar analysis is possible for topic subscriptions via [Filter](https://rfc.vac.dev/waku/standards/core/12/filter) protocol as well.
```
jm-clius, line 21: It is better not to tie specifications to a temporary state of the art as with "as of today". I'd rather say something like: "without a further anonymisation layer, it is easy..."

standards/core/mix.md
```diff
@@ -0,0 +1,119 @@
...
+
+1. any node can generate a mix packet and publish into the mix network and there needs to be some validation as to who is allowed to publish and whether the user is within allowed rate-limits.
+2. any node can generate paths which are broken intentionally and generate messages into the network.
+3. prevent a single attacker from spawning a huge number of mix nodes so that the user can observe and determine traffic patterns in order to deanonymize users. 
+
+The current [WAKU RLN](https://rfc.vac.dev/waku/standards/core/17/rln-relay) would not be enough to handle scenarios 1 and 2 explained above. RLN can address scenario-3 by making it restrictive for users to be able to spawn a lot of nodes in the network.
+
+We need to look at either customizing RLN or some alternative mechanism in order go enforce rate-limits or spam protect the mix network.
+
+## Potential problems (Need better name for the section)
```
jm-clius, line 94: Perhaps this can be called "Tradeoffs"?

standards/core/mix.md
```diff
@@ -0,0 +1,119 @@
...
+
+Combining existing discovery mechanisms of [WAKU DISCV5](https://rfc.vac.dev/waku/standards/core/33/discv5), [WAKU PEER-EXCHANGE](https://rfc.vac.dev/waku/standards/core/34/peer-exchange) and [WAKU RENDEZVOUS](https://github.com/waku-org/specs/blob/master/standards/core/rendezvous.md) should help address this problem till a certain extent, but not completely. 
+
+As Waku rendezvous proposes to use a 10 second registration interval and a 1 minute expiry for all nodes to a rendezvous point, this should help nodes maintain up-to-date online status of other mix nodes in the network.
+
+Edge nodes will have to use rendezvous discovery in order to discover a larger and available set of mix nodes quickly.
+
+Having following improvements in discovery would further improve the efficacy of mix protocol:
+- quicker discovery of most of the mix nodes
+- to be aware of online status of the mix nodes.
```
jm-clius, line 80: I think it's OK to have a "Discovery" section in the raw spec where we state the problem (as you've done above), but I would remove any of the existing implementation details (i.e. line 72 onwards). It's OK to leave unknown sections unspecified, rather than reporting on existing shortcomings (which would date the spec quickly)

standards/core/mix.md
```diff
@@ -0,0 +1,119 @@
...
+
+Light/Edge nodes will only be acting as `entry` nodes.
+
+## Flow
+
+TBD
+
+## ENR updates
+
+Each waku node that supports mix shall include the same in its ENR as per [Waku ENR](https://github.com/waku-org/specs/blob/master/standards/core/enr.md).
```
jm-clius, line 58: Feeling ambivalent as to whether we should include anything on the ENRs at all. Nevertheless, I think this is OK to include for now, but I would opt for more precise/straightforward "spec language" where there is no room for ambiguity. It sounds a bit halting and staccato to read, but helps with clarity. Something like:

Each Waku node that supports mix SHOULD indicate this in its discoverable ENR.
` bit 5` in the ENR...is reserved to indicate `mix` support.
This bit MUST be set to `true` to indicate mix support.

standards/core/mix.md
```diff
@@ -0,0 +1,119 @@
...
+Nodes shall advertise that they support mix protocol via their [ENR](#ENR updates).
+Nodes that would like to use mix service shall discover enough mix nodes so that there is always a healthy pool of mix nodes available for selection. The pool size of mix nodes should be in multiples of 100 for the mixing to be effective.
+Nodes that want higher anonymity while publishing a message or performing a store query would use the mix protocol to route their messages to their destination.
+A response/reply to a protocol request sent using Mix can be implemented using `Single Use reply blocks` as the service node will not be aware of the originator of the request.
+In order to have better reliability, any node that wish to send message using mix shall use 2 redundant paths.
+
+## Terminology
+The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”, “SHOULD”, “SHOULD NOT”, 
+“RECOMMENDED”, “MAY”, and “OPTIONAL” in this document are to be interpreted as described in [RFC2119](https://www.ietf.org/rfc/rfc2119.txt).
```
jm-clius, line 38: I'd move this to before the Specification section

standards/core/mix.md
```diff
@@ -0,0 +1,119 @@
...
+
+
+# Specification
+
+Mix protocol integration with waku would end up creating a separate mix network of all the waku nodes that also support the mix protocol. All mix packets would be routed via this mix network before reaching their intended destination.
+Nodes shall advertise that they support mix protocol via their [ENR](#ENR updates).
+Nodes that would like to use mix service shall discover enough mix nodes so that there is always a healthy pool of mix nodes available for selection. The pool size of mix nodes should be in multiples of 100 for the mixing to be effective.
+Nodes that want higher anonymity while publishing a message or performing a store query would use the mix protocol to route their messages to their destination.
+A response/reply to a protocol request sent using Mix can be implemented using `Single Use reply blocks` as the service node will not be aware of the originator of the request.
+In order to have better reliability, any node that wish to send message using mix shall use 2 redundant paths.
```
jm-clius, line 34: I'd consider phrasing these with the following sentence template in mind

<Explicit agent> MUST/SHOULD/MAY <perform explicit action> (optional: to explicit object/under explicit condition/etc.)

In other words, use the RFC keywords as guideline to think about declarative statements that an implementer of Waku protocols can follow as a "recipe" to decide how and where to include mix. Example suggestions:

Nodes with `mix` protocol mounted SHOULD advertise that they support mix via their chosen discovery method.
They MAY do so by updating....their ENR according to section...and using one of the ENR-based discovery methods, e.g....
Nodes that would like to use `mix`, SHOULD discover enough mix nodes...
The pool size SHOULD be larger than...

jm-clius (commented):
A couple of suggestions below. Generally I feel the structure is good, but I'd like to see a bit more of a descriptive heuristic that can be followed as a "recipe" in order to introduce `mix` into Waku.

standards/core/mix.md
chaitanyaprem: got it, will change

standards/core/mix.md
chaitanyaprem: I was also having a doubt if i should explain it here or update the ENR spec and refer the updated section. 
Will include it for now, we can move it later and update the ENR spec once this formalizes little bit. 

Understood, i guess i was not clear as to level of detail we need to write in a spec. Now i get that we only need to state what we are clearly thinking even if that may change in subsequent phase.

jm-clius:
Requesting some initial input from @jimstir and Vac RFC team too.

standards/core/mix.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Terminology
+The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”, “SHOULD”, “SHOULD NOT”, 
+“RECOMMENDED”, “MAY”, and “OPTIONAL” in this document are to be interpreted as described in [RFC2119](https://www.ietf.org/rfc/rfc2119.txt).
+
+## Node Roles
+
+Mix protocol defines 3 roles for the nodes in the mix network - `entry`, `exit`, `intermediary`. 
+Each node in the network can have all 3 roles or can act as just an entry/exit node as well. 
+
```
jimstir, line 57: When introducing node roles, keywords should be used. For example:
A Waku node SHOULD implement all 3 roles or SHOULD only implement an entry and exit node.

standards/core/mix.md
```diff
@@ -0,0 +1,139 @@
...
+- An `entry` node is the originator node of a message, i.e a node that wishes to publish/query messages to/from the waku network.
+- An `exit` node is responsible for delivery messages to destination peer in the network.
+- An `intermediary` node is responsible for forwarding a mix packet to the next mix node in the path.
+
+All the relay nodes would by default act as mix `intermediary` and `exit` nodes in the network.
```
jimstir, line 62: This statement contradicts the previous statement: "Each node in the network can have all 3 roles or can act as just an entry/exit node as well." The previous statement implies that there are only two possible node implementations but the next statement introduced another node implementation. I suggest moving the previous statement here and clearly describe all the possible node roles a relay/Waku node can act as in the same place. Also when using the term network, be specific of the network each time (Waku network or Waku mix).

jimstir (commented):
Hey!! I know this is a work in progress. I left a few pointers that you can use when you continue to work on this spec. These are just suggestions, so feel free to use anything useful.

standards/core/mix.md
chaitanyaprem: I will rephrase the statements, the first statement "Each node..." is meant to just indicate that any mix node role can be supported by any node..but looks like it is not clearly stated. 

These current statments show what type of waku node supports which mix node roles. I will rephrase them to make it clear.

standards/core/mix.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Background / Rationale / Motivation
+
+Waku protocols have weak sender/originator anonymity as explained in [Waku Privacy and Anonymity Analysis](https://vac.dev/rlog/wakuv2-relay-anon/).
+Without further anonymization, it is easy to analyze the network traffic and determine originator of messages published into the network or determine the topic interests by analyzing store queries requested by a user. 
+Similar analysis is possible for topic subscriptions via [Filter](https://rfc.vac.dev/waku/standards/core/12/filter) protocol as well.
+
+By using the libp2p mix protocol, we can improve the anonymity for publishers and store query users in the waku network.
+Anonymity of Filter users is not be addressed by this document and will have to be addressed separately.
+
```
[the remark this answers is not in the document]
chaitanyaprem: I am not thinking this is a protocol spec as it is describing mix integration into waku. 
Not sure where exactly this falls, but for now just looking at it as a spec that describes integrating mix protocol with waku.

standards/core/mix.md
```diff
@@ -0,0 +1,125 @@
...
+Waku Mix creates an `overlay network` of all the waku nodes that support the `mix` protocol.
+
+Nodes with `mix` protocol mounted SHOULD advertise that they support mix protocol via their their chosen discovery method.
+They may do so by updating their [ENR](#enr-updates) and using one of the ENR based discovery methods.
+
+Nodes that want higher anonymity while publishing a message via [lightpush](https://rfc.vac.dev/waku/standards/core/19/lightpush) or performing a [store](https://rfc.vac.dev/waku/standards/core/13/store) query SHOULD use the `mix` protocol to route their messages to the destination.
+Sender Node that would like to use `mix` protocol SHOULD discover enough mix nodes so that there is always a healthy pool of mix nodes available for selection.
+The pool size of mix nodes SHOULD be large enough(greater than 100) for the mixing to be effective.
+
+In order to provide anonymity of the sender node from the `service-node`,  `Single Use reply blocks` or `anonymous replies` as specified in the original [sphinx](#references) paper SHALL be used.
```
SionoiS, line 56: Would it be pertinent to add a few words about how a Waku protocol payload is encapsulated by sphinx payload?

standards/core/mix.md
```diff
@@ -0,0 +1,125 @@
...
+
+A node that sends messages using `mix` MAY use 2 redundant (paths)[https://rfc.vac.dev/vac/raw/mix#24-node-discovery] to have better reliability of the message being delivered.
```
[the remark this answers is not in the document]
chaitanyaprem: Redundant requests are not part of sphinx packet format. This is something we are thinking of doing at user-layer of mix to improve reliability. 
The paths chosen would be different for the redundant requests so the sphinx packet itself would be different and hence relayed accordingly via the path chosen for each request. 
Only the sender and destination node will know that the underlying payload in sphinx packet is the same, at the mix protocol layer these would act as individual packets.

standards/core/mix.md
chaitanyaprem: This shouldn't be part of this spec rather already explained in libp2p-mix spec here https://rfc.vac.dev/vac/raw/mix#4-sphinx-packet-format. 
Waku packet would just be the payload in the sphinx packet. 
Now that you mention it, would be a good idea to indicate this point explicitly. Will add a point to convey the same. Thanks!

standards/core/mix.md
```diff
@@ -0,0 +1,125 @@
...
+
```
jm-clius, line 59: Even if it feels "obvious" I think it's a good idea to include a Wire Protocol section that specifies how the Waku protocols are encapsulated into libp2p mix - presumable for both Lightpush and Store the entire request is serialised and populated to a payload field?

jm-clius (approved):
Great work! 🙌
My comments below are either nitpicks or things that can be addressed in future PRs. I think this is good to merge as raw spec. :D

standards/core/mix.md
```diff
@@ -0,0 +1,125 @@
...
+## Spam Protection
+
+Mix protocol in waku network SHOULD have `rate-limiting/spam` protection to handle scenarios such as below:
+
+1. Any node can generate a mix packet and publish into the mix network. Hence there needs to be some validation as to who is allowed to publish and whether the user is within allowed rate-limits.
+2. Any node can intentionally generate paths which are broken and send messages into the mix network.
+3. An attacker can spawn a huge number of mix nodes so that user behaviour is observed in order to determine traffic patterns and deanonymize users. 
+
+The current [WAKU RLN](https://rfc.vac.dev/waku/standards/core/17/rln-relay) can address scenario-3 by making it restrictive for users to be able to spawn a lot of nodes in the network.
```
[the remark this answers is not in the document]
chaitanyaprem: Thanks, you are right that RLN is only required for publishing and not for relaying, so any user can still spawn a lot of waku nodes and observe the network and deduce or deanonymize. Will remove this point.

standards/core/mix.md
chaitanyaprem: Yep, i am adding a line indicating that `Waku-Message` would become the payload field in the sphinx packet format.

standards/core/mix.md
```diff
@@ -0,0 +1,128 @@
...
+A Waku relay node SHOULD by default have mix `intermediary` and `exit` node roles in the network.
+The implementation MAY provide a configuration to disable a node from acting as an `intermediary\exit` node.
+
+Any waku node that wishes to publish/query messages from the waku network MUST act as a `sender` node.
+
+Resource-restricted/Edge nodes with short connection windows SHOULD always be acting as `sender` nodes.
+
+## ENR updates
+
+Each waku node that supports `mix` SHOULD indicate the same in its discoverable [ENR](https://github.com/waku-org/specs/blob/master/standards/core/enr.md).
```
fryorcraken, line 81: The way it's phrased, then a `sender` node "supports `mix`" and hence would indicate it in the ENR. I don't think it's correct.

I would only ask `intermediary` and `exit` nodes (ie, those part of the path) to advertise supporting `mix`.

standards/core/mix.md
```diff
@@ -0,0 +1,128 @@
...
+The following fields MUST be set as part of the discoverable ENR of a mix waku node:
+- The `bit 5` in the [waku2 ENR key](https://github.com/waku-org/specs/blob/master/standards/core/enr.md#waku2-enr-key) is reserved to indicate `mix`support. This bit MUST be set to true to indicate `mix` support.
+- A new field `mix-key` SHOULD be set to the `ed25519 public key` which is used for sphinx encryption.
+
+### Discovery
+
+Mix protocol provides better anonymity when an originator node has a sufficiently large pool of mix nodes to do path selection.
+This moves the problem into discovery domain and requires the following from discovery mechanisms:
+1. It is important for nodes to be able to discover as many nodes as possible quickly. This becomes especially important for edge nodes that come online just to publish/query messages for a short period of time.
+2. It is important to have the most recent online status of the nodes so that mix paths that are selected are not broken which lead to reliability issues.
```
fryorcraken, line 91: Would that be the reason we add a TTL or timestamp to ENRs? :thinking:

standards/core/mix.md
```diff
@@ -0,0 +1,128 @@
...
+## Spam Protection
+
+Mix protocol in waku network SHOULD have `rate-limiting/spam` protection to handle scenarios such as below:
+
+1. Any node can generate a mix packet and publish into the mix network. Hence there needs to be some validation as to who is allowed to publish and whether the user is within allowed rate-limits.
+2. Any node can intentionally generate paths which are broken and send messages into the mix network.
+3. An attacker can spawn a huge number of mix nodes so that user behaviour is observed in order to determine traffic patterns and deanonymize users. 
+
+There is a need to enforce rate-limits and spam protect the mix network. 
+The rate-limiting and spam protection SHALL be addressed as part of future work.
```
fryorcraken, line 105:
```suggestion
The rate-limiting and spam protection shall be addressed as part of future work.
```

It's not a protocol/behaviour statement so I don't think RFC-2119 applies

standards/core/mix.md
```diff
@@ -0,0 +1,128 @@
...
+Any waku node that wishes to publish/query messages from the waku network MUST act as a `sender` node.
+
+Resource-restricted/Edge nodes with short connection windows SHOULD always be acting as `sender` nodes.
+
+## ENR updates
+
+Each waku node that supports `mix` SHOULD indicate the same in its discoverable [ENR](https://github.com/waku-org/specs/blob/master/standards/core/enr.md).
+The following fields MUST be set as part of the discoverable ENR of a mix waku node:
+- The `bit 5` in the [waku2 ENR key](https://github.com/waku-org/specs/blob/master/standards/core/enr.md#waku2-enr-key) is reserved to indicate `mix`support. This bit MUST be set to true to indicate `mix` support.
+- A new field `mix-key` SHOULD be set to the `ed25519 public key` which is used for sphinx encryption.
```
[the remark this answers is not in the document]
chaitanyaprem: Yeah i know we may hit ENR size limits, but to begin with this is the simpler way to go ahead. We will have to think of some other approach to address this size limit. I think keysize is 64 bytes.

standards/core/mix.md
chaitanyaprem: You are right, will modify it.

standards/core/mix.md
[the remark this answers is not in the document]
chaitanyaprem: Maybe this can be communicated/fetched using Kademelia which i think is in plan to replace discv5.

standards/core/mix.md
```diff
@@ -0,0 +1,128 @@
...
+Waku protocols have weak sender/originator anonymity as explained in [Waku Privacy and Anonymity Analysis](https://vac.dev/rlog/wakuv2-relay-anon/).
+Without further anonymization, it is easy to analyze the network traffic and determine the originator of messages published into the network.
+Topic interests of a user can be identified by analyzing [Store](https://rfc.vac.dev/waku/standards/core/13/store) query messages sent by a user.
+Same applies for topic subscriptions via [Filter](https://rfc.vac.dev/waku/standards/core/12/filter) protocol.
+
+`Mix` protocol allows libp2p nodes to send messages without revealing the sender's identity (peer ID, IP address) to intermediary mix nodes or the recipient/destination.
+Anonymity is achieved by using the [Sphinx packet format](#references), which encrypts and routes messages through a series of mix nodes before reaching the recipient.
+
+By integrating the `mix` protocol into the waku network, we can improve the anonymity for publishers and store query users.
+Each waku relay node will be acting as a mix node that forms an `overlay mix network`.
```
s-tikhomirov, line 36: "will be acting" is a bit unclear - how mandatory is this? Should it be clarified with MUST / SHALL / SHOULD?

standards/core/mix.md
```diff
@@ -0,0 +1,128 @@
...
+
+Anonymity of [Filter](https://rfc.vac.dev/waku/standards/core/12/filter) users is not addressed by this document.
+
+## Terminology
+The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”, “SHOULD”, “SHOULD NOT”,
+“RECOMMENDED”, “MAY”, and “OPTIONAL” in this document are to be interpreted as described in [RFC2119](https://www.ietf.org/rfc/rfc2119.txt).
+
+## Theory / Semantics
+
+Waku Mix creates an `overlay network` of all the waku nodes that support the `mix` protocol.
```
s-tikhomirov, line 47: Nitpick: capitalize "Waku"?

standards/core/mix.md
```diff
@@ -0,0 +1,128 @@
...
+
+Nodes with `mix` protocol mounted SHOULD advertise that they support mix protocol via their their chosen discovery method.
+They MAY do so by updating their [ENR](#enr-updates) and using one of the ENR based discovery methods.
+
+Nodes that want higher anonymity while publishing a message via [lightpush](https://rfc.vac.dev/waku/standards/core/19/lightpush) or performing a [store](https://rfc.vac.dev/waku/standards/core/13/store) query SHOULD use the `mix` protocol to route their messages to the destination.
+Sender Node that would like to use `mix` protocol SHOULD discover enough mix nodes so that there is always a healthy pool of mix nodes available for selection.
+The pool size of mix nodes SHOULD be large enough for the mixing to be effective.
+We RECOMMEND a pool size of at least 100 mix nodes for the mixing to be effective.
```
s-tikhomirov, line 55: Is there any calculation to back up this number? What is considered "effective", by what metric?

standards/core/mix.md
```diff
@@ -0,0 +1,128 @@
...
+
+The serialized [Waku Message](https://rfc.vac.dev/waku/standards/core/14/message) MUST be the payload in the [sphinx packet](https://rfc.vac.dev/vac/raw/mix#4-sphinx-packet-format).
+
+In order to provide anonymity of the sender node from the `service-node`,  `Single Use reply blocks` or `anonymous replies` as specified in the original [sphinx](#references) paper SHALL be used.
+
+A node that sends messages using `mix` MAY use two redundant (paths)[https://rfc.vac.dev/vac/raw/mix#24-node-discovery] to have better reliability of the message being delivered.
```
s-tikhomirov, line 61: What does it mean for two paths to be redundant? They must not have any node in common apart from the destination?

standards/core/mix.md
```diff
@@ -0,0 +1,128 @@
...
+It is up to the higher-layer mixed protocol to deduplicate redundant messages received in this way.
+
+## Node Roles
+
+Mix protocol defines 3 roles for the nodes in the mix network - `sender`, `exit`, `intermediary`.
+
+- An `sender` node is the originator node of a message, i.e a node that wishes to publish/query messages to/from the waku network.
+- An `exit` node is responsible for delivering messages to destination peer in the network.
```
s-tikhomirov, line 69: From what I read in the mix spec, by `exit` we mean the node that unpacks the inner-most Sphinx packet and supports the requested libp2p protocol - if so, then `exit` node _is_ the destination peer, right?

standards/core/mix.md
```diff
@@ -0,0 +1,128 @@
...
+- An `intermediary` node is responsible for forwarding a mix packet to the next mix node in the path.
+
+A Waku relay node SHOULD by default have mix `intermediary` and `exit` node roles in the network.
+The implementation MAY provide a configuration to disable a node from acting as an `intermediary\exit` node.
+
+Any waku node that wishes to publish/query messages from the waku network MUST act as a `sender` node.
```
s-tikhomirov, line 75: "that wishes to publish / query messages" _via `mix`_ is implied?

standards/core/mix.md
```diff
@@ -0,0 +1,128 @@
...
+## ENR updates
+
+Each waku node that supports the `mix intermediary or exit role` SHOULD indicate the same in its discoverable [ENR](https://github.com/waku-org/specs/blob/master/standards/core/enr.md).
+The following fields MUST be set as part of the discoverable ENR of a mix waku node:
+- The `bit 5` in the [waku2 ENR key](https://github.com/waku-org/specs/blob/master/standards/core/enr.md#waku2-enr-key) is reserved to indicate `mix` support. This bit MUST be set to true to indicate `mix` support.
+- A new field `mix-key` SHOULD be set to the `ed25519 public key` which is used for sphinx encryption.
+
+### Discovery
+
+Mix protocol provides better anonymity when an originator node has a sufficiently large pool of mix nodes to do path selection.
```
s-tikhomirov, line 88: Use "sender" instead of "originator" for consistency?

standards/core/mix.md
```diff
@@ -0,0 +1,128 @@
...
+
+Point-2 above can be mitigated partially by choosing redundant mix paths for the same message by the sender node.
+This may not be an effective solution as it increases the overall bandwidth usage.
+
+## Spam Protection
+
+Mix protocol in waku network SHOULD have `rate-limiting/spam` protection to handle scenarios such as below:
+
+1. Any node can generate a mix packet and publish into the mix network. Hence there needs to be some validation as to who is allowed to publish and whether the user is within allowed rate-limits.
+2. Any node can intentionally generate paths which are broken and send messages into the mix network.
```
s-tikhomirov, line 101: What is a broken path? Is it a path that includes a non-existent node? Other than that, do we assume that two live mix node can always form a hop in a path if a Sphinx packet says so?

s-tikhomirov (approved):

standards/core/mix.md
chaitanyaprem: I will change it to SHOULD.
Ideally we want all relay nodes to be part of the mix network so that node-pool is larger.
But that would require further analysis and may lead to incentivization if needed in future.

standards/core/mix.md
chaitanyaprem: Not necessarily, it just means choose 2 paths (each path is chosen randomly from node-pool). If the node pool is sufficiently large then the chance of a node being common in 2 paths selected in quick succession would be very low. But there is no gaurantee as such to ensure there is no common node in 2 paths.

standards/core/mix.md
chaitanyaprem: The mix spec needs to be updated as the implementation has been changed to remove this dependency that exit node has to support the requested libp2p protocol. As of now our POC also doesn't assume this.
You can follow the discussion [here](https://forum.vac.dev/t/introducing-the-mix-protocol-enhancing-privacy-across-libp2p-networks/348/10) for reference

standards/core/mix.md
chaitanyaprem: good catch, didn't mean to imply it..but i will add this.

standards/core/mix.md
chaitanyaprem: well, here it just means that somewhere along the path one of the mix nodes is not able to connect to the next node in the path. It could be that the node is no longer running or offline or just not reachable. 
Some of these can be mitigated if the sender of the messages checks rechability and online status of all nodes in the path before sending the packet out. 
These are all challenges that need to be solved in further phases.

standards/core/mix.md
```diff
@@ -0,0 +1,128 @@
...
+Same applies for topic subscriptions via [Filter](https://rfc.vac.dev/waku/standards/core/12/filter) protocol.
+
+`Mix` protocol allows libp2p nodes to send messages without revealing the sender's identity (peer ID, IP address) to intermediary mix nodes or the recipient/destination.
+Anonymity is achieved by using the [Sphinx packet format](#references), which encrypts and routes messages through a series of mix nodes before reaching the recipient.
+
+By integrating the `mix` protocol into the waku network, we can improve the anonymity for publishers and store query users.
+Each waku relay node will be acting as a mix node that forms an `overlay mix network`.
+This network of mix nodes SHALL relay mix messages anonymously to the recepient.
+
+Anonymity of [Filter](https://rfc.vac.dev/waku/standards/core/12/filter) users is not addressed by this document.
```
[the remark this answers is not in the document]
chaitanyaprem: Hmm, i just wanted to indicate high level scope here as to what is covered and not covered. Did not want to add a separate scope section. There is no such section as of now about things that are not covered. Maybe as we mature the rfc further we can refine it.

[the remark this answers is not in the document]
s-tikhomirov: I meant that this sentence in the abstract - "Integration of [waku relay](https://rfc.vac.dev/waku/standards/core/11/relay) with mix is not covered in this document." - also states that something is out of scope. My suggestion was to move all mentions of things that are out of scope to one place somewhere - I've no strict preference here though, just a suggestion.

jimstir (approved):
👌 good spec writing

standards/core/mix.md
fryorcraken: Would we want to use rendezvous in edge nodes? cc @SionoiS

standards/core/mix.md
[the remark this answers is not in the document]
fryorcraken: Maybe we just highlight waht limitations we are/may encounter with discv5/enr

standards/core/mix.md
SionoiS: > Would we want to use rendezvous in edge nodes? cc @SionoiS

Yes it make sense since it's very lively. It takes care of the discovery problem it's just not very decentralized.

standards/core/mix.md
```diff
@@ -0,0 +1,128 @@
...
+
+Each waku node that supports the `mix intermediary or exit role` SHOULD indicate the same in its discoverable [ENR](https://github.com/waku-org/specs/blob/master/standards/core/enr.md).
+The following fields MUST be set as part of the discoverable ENR of a mix waku node:
+- The `bit 5` in the [waku2 ENR key](https://github.com/waku-org/specs/blob/master/standards/core/enr.md#waku2-enr-key) is reserved to indicate `mix` support. This bit MUST be set to true to indicate `mix` support.
+- A new field `mix-key` SHOULD be set to the `ed25519 public key` which is used for sphinx encryption.
+
+### Discovery
+
+Mix protocol provides better anonymity when a sender node has a sufficiently large pool of mix nodes to do path selection.
+This moves the problem into discovery domain and requires the following from discovery mechanisms:
```
AkshayaMani, line 89: "_The discovery mechanism MUST be unbiased_"  should be explicitly stated as a requirement. If the discovery process favours certain subsets of nodes, adversaries could exploit this to increase the likelihood of compromised nodes being chosen. Ensuring unbiased discovery is a fundamental requirement for preserving the anonymity guarantees of the mix network.

standards/core/mix.md
```diff
@@ -0,0 +1,128 @@
...
+Point-2 above can be mitigated partially by choosing redundant mix paths for the same message by the sender node.
+This may not be an effective solution as it increases the overall bandwidth usage.
+
+## Spam Protection
+
+Mix protocol in waku network SHOULD have `rate-limiting/spam` protection to handle scenarios such as below:
+
+1. Any node can generate a mix packet and publish into the mix network. Hence there needs to be some validation as to who is allowed to publish and whether the user is within allowed rate-limits.
+2. Any node can intentionally generate paths which are broken and send messages into the mix network.
+3. An attacker can spawn a huge number of mix nodes so that user behaviour is observed in order to determine traffic patterns and deanonymize users.
```
AkshayaMani, line 102: It may be worth clarifying that while an attacker can introduce a large number of mix nodes to analyze traffic patterns, Sphinx encryption makes such attacks significantly more difficult. An adversary must control at least both the entry and exit nodes to successfully deanonymize a user, which becomes increasingly difficult as the network size grows.

standards/core/mix.md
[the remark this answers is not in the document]
AkshayaMani: > Yeah i know we may hit ENR size limits, but to begin with this is the simpler way to go ahead. We will have to think of some other approach to address this size limit. I think keysize is 64 bytes.
Just a quick clarification—the key size is 32 bytes, not 64. That said, hitting ENR size limits is a valid concern and requires further consideration.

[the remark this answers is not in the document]
chaitanyaprem: > Just a quick clarification—the key size is 32 bytes, not 64. That said, hitting ENR size limits is a valid concern and requires further consideration.

Thanks, i might have checked privateKey size :)

standards/core/mix.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Theory / Semantics
+
+Waku Mix creates an `overlay network` of all the Waku nodes that support the `mix` protocol.
+
+Nodes with `mix` protocol mounted SHOULD advertise that they support `mix` protocol via their their chosen discovery method.
+They MAY do so by updating their [ENR](#enr-updates) and using one of the ENR based discovery methods.
+
+Nodes that want higher anonymity while publishing a message via [lightpush](https://rfc.vac.dev/waku/standards/core/19/lightpush) or performing a [store](https://rfc.vac.dev/waku/standards/core/13/store) query SHOULD use the `mix` protocol to route their messages to the destination.
+Sender Node that would like to use `mix` protocol SHOULD discover enough mix nodes so that there is always a healthy pool of mix nodes available for selection.
```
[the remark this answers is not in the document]
chaitanyaprem: ok, that is a good point i.e discovery should not choose nodes that are topologically closer and should be unbiased. Will add that explicitly.
 
100 was just a number i have mentioned considering network size of 1000(which is the approx size of The Waku Network), did not arrive at it based on specific logic but more of a hunch :) 

> Therefore, explicitly stating these assumptions could be helpful—the discovered mix pool MUST be unbiased, and the network size `N` MUST be sufficiently large compared to the pool size.

wondering if we should state the network size assumption. What if network size is itself 100, then having nodepool size of 100 also is good imo.  Can we rather mention that node pool size should be atleast x% of network size or 100 whichever is applicable, not sure what `x` should be?

standards/core/mix.md
chaitanyaprem: understood, but if the network makes it easy to spawn nodes (i.e at low cost and no friction of entry) then an attacker with resources can spin up  a lot of nodes very easily.

This has to be solved some other way, which is why i wanted to highlight this point explicitly.

PR #55, opened 2025-02-26
Incremental spellcheck action

jazzz:
## Summary

This PR updates the Github action used to preform spellchecks on commits. 

## Problem
Currently the workflows are [failing](https://github.com/waku-org/specs/actions) for every PR due to spelling "errors" in existing markdown files. These errors are mostly valid words which have not been added to the dictionary. 
- It is difficult to tell if the PR contains new spelling errors or not, as the workflow is consistently failing.
- Fixing all the errors at once would be chaotic and time consuming.

## Solution
Change the approach to spellcheck.

- On PullRequest: spell check modified files
    - Ensures new commits are error free.
    - Promotes fixing specifications that are being actively updated
- Once a week: spellcheck the entire repo
    - Provides some pressure towards cleaning up existing specifications, without adding significant friction.
    
This results in usable spellcheck for new additions, while not ignoring potential existing errors.

## Notes
- Currently the Spellcheck(Incremental) workflow is run on the whole repo, when there are commits with no modified markdown files.  
    - This doesn't seem terrible for a specifications repository.
    - Of course that means that this PR will fail, as no markdown files were modified
- The existing action was old and un maintained, so it was updated to a more active project (which also supports incremental checks)
- Wordlist was separated from the configuration file, to increase hygiene.


jm-clius (commented):
Thanks for this! Indeed,, the failing CI is something to fix. I like this approach, but have requested a review from @jimstir in the RFC team to consider if this is a good solution across org-wide spec repos.

jimstir:
This approach is an improvement. When having a spell-check workflow the wordlist.txt will need to be maintained to have no failures for each pr ( which is doable ). Earlier last year we discussed adding spell-check but decided to just have linting. If necessary could open discussion to be added later.

jimstir (approved):

PR #56, opened 2025-02-26
docs: fix TeX superscript

jazzz:
This PR fixes the rendering of exponents in `relay-sharing.md`. TeX requires braces be used when the superscript applied to multiple characters.

**Previous**
image

**Changed**
image

## Notes

- While not critical, this does help with readability in HTML. 
- Plaintext readability is not significantly affected


jm-clius (approved):
Thanks

PR #57, opened 2025-03-03
chore: waku sync full topic support

SionoiS:
update to the Waku sync spec to include full topic support.

### DIFF
--- a/standards/core/sync.md
+++ b/standards/core/sync.md
@@ -36,7 +36,7 @@ At this point lists of message _IDs_ are sent for comparison
 instead of _fingerprints_ over entire ranges of messages.
 
 #### Overview
-The `reconciliation` protocol operate on the following heuristic:
+The `reconciliation` protocol operates on the following heuristic:
 1. The requestor chooses a sync range including time, cluster id, pubsub and content topics.
 2. The range is encoded into a payload and sent.
 3. The requestee receives the payload and decodes it.
--- a/standards/core/sync.md
+++ b/standards/core/sync.md
@@ -37,7 +37,7 @@ instead of _fingerprints_ over entire ranges of messages.
 
 #### Overview
 The `reconciliation` protocol operate on the following heuristic:
-1. The requestor chooses a sync range including time, cluster id and topics.
+1. The requestor chooses a sync range including time, cluster id, pubsub and content topics.
 2. The range is encoded into a payload and sent.
 3. The requestee receives the payload and decodes it.
 4. The range is processed and, if a difference with the local range is detected, a set of subranges are produced.
@@ -76,10 +76,10 @@ Every _range_ MUST have one of the following types; _fingerprint_, _skip_ or _it
 recursing further reduce the number of round-trips.
 
 #### Sync Scope
-On payload reception, the intersection of the two peers topic sets is used as the sync scope.
+On payload reception, the intersection of the two peers pubsub and content topic sets is used as the sync scope.
 If the intersection is empty the sync is aborted.
-If a peer does not specify a set of supported topics,
-the protocol assumes all topics are supproted for that peer.
+If a peer does not specify a set of supported pubsub or content topics,
+the protocol assumes all pubsub or content topics are supported for that peer.
 
 #### Range Processing
 _Ranges_ have to be processed differently according to their types.
@@ -188,8 +188,8 @@ What will follow is NOT part of the specification.
 This section was created to inform implementations.
 
 #### Cluster, Pubsub and Content Topics
-To prevent nodes from synchronizing messages from cluster and topics they don't support,
-cluster and topics information is added to each payload.
+To prevent nodes from synchronizing messages from cluster, pubsub and content topics they don't support,
+cluster and topic information is added to each payload.
 
 #### Parameters 
 Two useful parameters to add to your implementation are partitioning count and the item set threshold.
--- a/standards/core/sync.md
+++ b/standards/core/sync.md
@@ -36,8 +36,8 @@ At this point lists of message _IDs_ are sent for comparison
 instead of _fingerprints_ over entire ranges of messages.
 
 #### Overview
-The `reconciliation` protocol operates on the following heuristic:
-1. The requestor chooses a sync range including time, cluster id, pubsub and content topics.
+The `reconciliation` protocol operate on the following heuristic:
+1. The requestor chooses a sync range including time, id, pubsub and content topics.
 2. The range is encoded into a payload and sent.
 3. The requestee receives the payload and decodes it.
 4. The range is processed and, if a difference with the local range is detected, a set of subranges are produced.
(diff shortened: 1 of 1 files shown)


standards/core/sync.md
```diff
@@ -72,6 +75,12 @@ Every _range_ MUST have one of the following types; _fingerprint_, _skip_ or _it
 > _Item sets_ are an optimization, sending multiple _IDs_ instead of
 recursing further reduce the number of round-trips.
 
+#### Sync Scope
+On payload reception, the intersection of the two peers topic sets is used as the sync scope.
+If the intersection is empty the sync is aborted.
+If a peer does not specify a set of supported topics,
+the protocol assumes all topics are supproted for that peer.
```
jm-clius, line 82:
```suggestion
the protocol assumes all topics are supported for that peer.
```

standards/core/sync.md
```diff
@@ -28,13 +28,16 @@ The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL
...
 This process repeats until _ranges_ include a small number of messages.
-At this point lists of message _IDs_ are sent for comparison instead of _fingerprints_ over entire ranges of messages.
+At this point lists of message _IDs_ are sent for comparison
+instead of _fingerprints_ over entire ranges of messages.
 
 #### Overview
-The `reconciliation` protocol follows the following heuristic:
-1. The requestor chooses a time range to sync.
+The `reconciliation` protocol operate on the following heuristic:
+1. The requestor chooses a sync range including time, cluster id and topics.
```
jm-clius, line 40: Presumably, with "topics" here you refer to both pubsub topics and content topics? We should be explicit about this here and elsewhere, but I wonder if we keep the spec limited to pubsub topics (only)? Even if in implementation we add experimental content topics support, my guess is that we'll mature/test the pubsub topics implementation before validating that it works for content topics. At that point, the spec can be upgraded to `draft`, while a new raw version can include content topics. If you agree, the suggested change then is to be explicit about "pubsub topic" throughout.

jm-clius (approved):
LGTM. Comment to address before merging below - we have to be explicit about "topic" as being pubsub topic, content topic or both. Suggestion is to limit spec only to behaviour for pubsub topic, which will allow it to undergo full testing and QA separately to reach `draft` status while we experiment with content topics in parallel (which IIUC may necessitate more comprehensive storage changes).

standards/core/sync.md
SionoiS: forgot the spell checker...

standards/core/sync.md
SionoiS: The mechanism is the same for pubsub or content topics but I agree about being explicit.

SionoiS: > Presumably, with "topics" here you refer to both pubsub topics and content topics?

I agree being explicit is better.

This PR reflects the changes in this https://github.com/waku-org/nwaku/pull/3275
Both should be merged together ounce we have tested the current version (which include shard support).

Since both pubsub and content topic use the exact same mechanism testing separately doesn't make much sense.

standards/core/sync.md
```diff
@@ -28,13 +28,16 @@ The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL
...
+the range _fingerprints_ are different, split into smaller (sub)ranges.
 This process repeats until _ranges_ include a small number of messages.
-At this point lists of message _IDs_ are sent for comparison instead of _fingerprints_ over entire ranges of messages.
+At this point lists of message _IDs_ are sent for comparison
+instead of _fingerprints_ over entire ranges of messages.
 
 #### Overview
-The `reconciliation` protocol follows the following heuristic:
-1. The requestor chooses a time range to sync.
+The `reconciliation` protocol operate on the following heuristic:
```
Ivansete-status, line 39:
```suggestion
The `reconciliation` protocol operates on the following heuristic:
```

standards/core/sync.md
```diff
@@ -28,13 +28,16 @@ The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL
...
+1. The requestor chooses a sync range including time, cluster id, pubsub and content topics.
```
Ivansete-status, line 40: Should we avoid the use of term pubsub?
```suggestion
1. The requestor chooses a sync range including time, cluster id, shard and content topics.
```

standards/core/sync.md
```diff
@@ -72,6 +75,12 @@ Every _range_ MUST have one of the following types; _fingerprint_, _skip_ or _it
 > _Item sets_ are an optimization, sending multiple _IDs_ instead of
 recursing further reduce the number of round-trips.
 
+#### Sync Scope
+On payload reception, the intersection of the two peers pubsub and content topic sets is used as the sync scope.
+If the intersection is empty the sync is aborted.
+If a peer does not specify a set of supported pubsub or content topics,
+the protocol assumes all pubsub or content topics are supported for that peer.
```
Ivansete-status, line 82: As I mentioned above, I think we should avoid the `pubsub` term. Instead, only `cluster-id`, `shard(s)`.
I'm suggesting that because the `pubsub` is composed by `cluster-d` and `shard`.

Ivansete-status (commented):
Thanks for it!
I think we should avoid using the term `pubsub` outside the nwaku codebase.
From API pov, I'd only use these terms:
- timestamp
- cluster-id
- shard
- content-topic

SionoiS:
> Thanks for it! I think we should avoid using the term `pubsub` outside the nwaku codebase. From API pov, I'd only use these terms:
> 
> * timestamp
> * cluster-id
> * shard
> * content-topic

I don't have strong opinions here we can either;
- use pubsub and content topics
- use cluster, shard and content topics

I think @jm-clius want the spec to be agnostic to network definitions?

jm-clius:
> I don't have strong opinions here we can either;
> 
> * use pubsub and content topics
> * use cluster, shard and content topics
> 
> I think @jm-clius want the spec to be agnostic to network definitions?

Most of our _core_ level specs use `pubsub_topic` as the mapping to pubsub topic from cluster ID and shard is a higher-level function. That said, I'm OK with use moving to cluster ID + shard throughout this spec, as long as we're consistent and do not refer to any specific deployment (e.g. TWN)

SionoiS:
> Most of our _core_ level specs use `pubsub_topic` as the mapping to pubsub topic from cluster ID and shard is a higher-level function. That said, I'm OK with use moving to cluster ID + shard throughout this spec, as long as we're consistent and do not refer to any specific deployment (e.g. TWN)

I'm lazy and I don't want to change it back to shard again so I'll leave it like this.

SionoiS:
I removed cluster and left only pubsub and content topics

igor-sirotin:
@SionoiS @Ivansete-status @jm-clius is this still relevant?

SionoiS:
> @SionoiS @Ivansete-status @jm-clius is this still relevant?

I guess not since waku sync is not used AFAIK.

Ivansete-status:
@igor-sirotin @SionoiS - store sync is currently disabled in both waku.sandbox and status fleets because it overloaded the DBs (IIRC) but I believe this is a very important feature. Enabling it again may require some minor tweaks so ensure better DB usage.

igor-sirotin:
I am marking this repository as archived now. 

@Ivansete-status if you want to get this spec changes in, please coordinate with AnonComms team, and open a new PR in [logos-lips](https://github.com/logos-co/logos-lips). Rebasing this PR here seems quite painful anyway.

cc @SionoiS @jm-clius

PR #64, opened 2025-06-05
docs: updates to the Store Sync and Waku Sync specs

jm-clius:
This PR proposes the following updates to the Store Sync and Waku Sync specifications:

1. For WAKU-SYNC, I've updated the Wire Protocol sections based on my understanding of the current implementation, especially how the encoding works. It would be great to check if my description is (i) accurate and (ii) complete (i.e. someone following only this spec can write a compatible WAKU-SYNC node that generates parseable payloads). Sidenote is that I think we can simplify the wire protocol in future iterations by packing all fields into `protobuf`s (while maintaining the differential encoding to compress payload size).
2. STORE-SYNC contains more or less the same essential information as before, but with some sections expanded.

@AYAHASSAN287, added you as reviewer simply to get your impression if this is sufficient for QA blackbox testing. No need for detailed review. I imagine that the `WAKU-SYNC` spec probably contains detail unnecessary for QA, so I'd recommend focusing on `STORE-SYNC`. Let us know if any specified behaviours are unclear. :)

### DIFF
--- a/standards/core/store-sync.md
+++ b/standards/core/store-sync.md
@@ -7,38 +7,111 @@ contributors:
 
 ## Abstract
 
-This document describe the strategy Waku Store node will employ to stay synchronized.
-The goal being that all store nodes eventually archive the same set of messages.
+This document describes a strategy to keep [13/WAKU2-STORE](https://rfc.vac.dev/waku/standards/core/13/store) nodes synchronised,
+using a combination of [13/WAKU2-STORE](https://rfc.vac.dev/waku/standards/core/13/store) queries 
+and the [WAKU-SYNC](../core/sync.md) protocol.
 
 ## Background / Rationale / Motivation
 
-Message propagation in the network is not perfect,
-even with GossipSub mechanisms to detect missed messages.
-Nodes can also go offline for various reason outside our control.
-Store nodes that want to provide a good service must be able to remedy situations like these.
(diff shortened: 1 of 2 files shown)


SionoiS (approved):
LGTM left some comments

jm-clius:
@Cofson requesting review here to help find gaps/inconsistencies in technical content. The Sync specs are fairly complex, but it's important that the wire protocol is self-sufficient (i.e. doesn't require knowledge of the nwaku implementation to develop a compatible implementation) and complete enough to allow QA to write blackbox tests.

AYAHASSAN287 (commented):
Thanks Hanno for all detailed explanation . I just have one comment , now reading the specs makes me understand the big picture of the protocol and the interaction between nodes but to write full black box scenario I need some details about the how it could be done .

1. example: 

`reconciliation MUST be used to detect and exchange differences between [14/WAKU2-MESSAGE](https://rfc.vac.dev/waku/standards/core/14/message)s cached by the [13/WAKU2-STORE](https://rfc.vac.dev/waku/standards/core/13/store) node`

how to enable reconciliation for node ? what flag shall be used ? same for transfer 

2.Example 
`Any peer selected for Store Sync SHOULD be chosen at random` 
what flags to assign peers for each node ?  

**conclusion :** 
For efficient and fast drafting of black box scenarios we require full set of options with description to be able to configure the node

PR #19, opened 2024-05-27
Add "Messaging API" spec

weboko:
This PR attempts to define `Messaging API` as per all previous conversations in various threads or offline. 

This doc describes following parts of said umbrella API:
- `initial configuration` - just configuration;
- `Send API` - send for message with retry and ack checks;
- `Subscribe API` - initiate, track and manage ongoing source of messages;
- `Messaging Storage API` - entry point for access for messages;
- ~`History API` - allows to fetch past messages;~
- `Event Source API` - provides reactive way for app devs;

These sections cover programmatic API as well as `REST` API. 

For code sections `typescript` was used AND I am not sure it is the best option for our case so I propose either to:
- add definitions in other languages alongside;
- change completely to pseudocode for definitions;

Unfortunately the document turned out rather big, with some degree of repetitions in explanations and references between sections. Please, let me know right after first read if it is too much and I will rewrite it for further review.

PR #65, opened 2025-06-17
Waku API Spec: Node initialization

fryorcraken:
This PR intends to introduce the customer Interface Definition Language, by proposing an initialization function and accompanying config for the Waku API.

The "Waku API" is a proposed renamed of "Waku Messaging API" because the term "messaging" is very generic and is going to also be used by Status to qualify the Waku Chat SDK API.

Also best to start generic for us with "Waku API", as this is the API we want developers to use. And we can refine at a later stage if there is a more specialized API we want to offer (aka "RLN API").

This PR is a suggestion to replace #19. I have already made comment in #19, but in short:

- More succinct document and definition, we are here to define an API, behaviour should be inferred by protocols already defined
- Focus on iterative delivery, the spec should match the code. eg we do not intend to implement Waku API in REST API in 2025H2, so the spec can be done later
- Trying to avoid duplicate block of text, and keeping definitions close to introduction
- Making it language agnostic, so it can truly define the API across all languages.

https://github.com/waku-org/specs/issues/68

### DIFF
--- a/standards/application/waku-api.md
+++ b/standards/application/waku-api.md
@@ -130,10 +130,6 @@ types:
       waku_config:
         type: WakuConfig
         default: TheWakuNetworkPreset
-      message_confirmation:
-        type: bool
-        default: true
-        description: "Whether to apply peer-to-peer reliability strategies to confirm that outgoing message have been received by other peers."
       networking_config:
         type: NetworkConfig
         default: DefaultNetworkingConfig 
@@ -281,18 +277,16 @@ values:
 
 **`mode`**:
 
-If the `mode` set is `edge`, the initialised `WakuNode` MUST mount:
+If the `mode` set is `edge`, the initialised `WakuNode` SHOULD use:
 
 - [LIGHTPUSH](https://github.com/vacp2p/rfc-index/blob/main/waku/standards/core/19/lightpush.md) as client 
 - [FILTER](https://github.com/vacp2p/rfc-index/blob/main/waku/standards/core/12/filter.md) as client
 - [STORE](https://github.com/vacp2p/rfc-index/blob/main/waku/standards/core/13/store.md) as client
 - [METADATA](https://github.com/vacp2p/rfc-index/blob/main/waku/standards/core/66/metadata.md) as client
+- [PEER-EXCHANGE](https://github.com/vacp2p/rfc-index/blob/main/waku/standards/core/34/peer-exchange.md) as client
+- [P2P-RELIABILITY](/standards/application/p2p-reliability.md)
 
-And must use mount and use the following protocols to discover peers:
-
-- [PEER-EXCHANGE](https://github.com/vacp2p/rfc-index/blob/main/waku/standards/core/34/peer-exchange.md)
-
-If the `mode` set is `core`, the initialised `WakuNode` MUST mount:
+If the `mode` set is `core`, the initialised `WakuNode` SHOULD use:
 
 - [RELAY](https://github.com/vacp2p/rfc-index/blob/main/waku/standards/core/11/relay.md)
 - [LIGHTPUSH](https://github.com/vacp2p/rfc-index/blob/main/waku/standards/core/19/lightpush.md) as service node
@@ -300,23 +294,13 @@ If the `mode` set is `core`, the initialised `WakuNode` MUST mount:
 - [PEER-EXCHANGE](https://github.com/vacp2p/rfc-index/blob/main/waku/standards/core/34/peer-exchange.md) as service node
 - [STORE](https://github.com/vacp2p/rfc-index/blob/main/waku/standards/core/13/store.md) as client
 - [METADATA](https://github.com/vacp2p/rfc-index/blob/main/waku/standards/core/66/metadata.md) as client and service node
-
-And must use mount and use the following protocols to discover peers:
-
+- [P2P-RELIABILITY](/standards/application/p2p-reliability.md)
 - [DISCV5](https://github.com/vacp2p/rfc-index/blob/main/waku/standards/core/33/discv5.md)
-- [PEER-EXCHANGE](https://github.com/vacp2p/rfc-index/blob/main/waku/standards/core/34/peer-exchange.md)
-- [RENDEZVOUS](https://github.com/waku-org/specs/blob/master/standards/core/rendezvous.md)
+- [PEER-EXCHANGE](https://github.com/vacp2p/rfc-index/blob/main/waku/standards/core/34/peer-exchange.md) as client and service node
+- [RENDEZVOUS](https://github.com/waku-org/specs/blob/master/standards/core/rendezvous.md) as client and service node
 
 `edge` mode SHOULD be used if node functions in resource restricted environment,
 whereas `core` SHOULD be used if node has no strong hardware or bandwidth restrictions.
-
-**`message_confirmation`**:
-
-As defined in [P2P-RELIABILITY](/standards/application/p2p-reliability.md).
-Proceed with confirmation on whether outgoing messages were received by other nodes in the network.
-
-When set to true, [Store-based reliability for publishing](/standards/application/p2p-reliability.md#1-store-based-reliability-for-publishing) SHOULD be enabled.
-In `edge` `mode`, [Retransmit on possible message loss detection](/standards/application/p2p-reliability.md#4-retransmit-on-possible-message-loss-detection) by installing filter subscription(s) matching the content topic(s) used for publishing, MAY be enabled.
 
 ## The Validation API
 


standards/application/waku-api.md
```diff
@@ -0,0 +1,201 @@
...
+Hence, instead of having the reader learn a new IDL, we propose to use a simple IDL with self-describing syntax.
+
+An alternative would be to choose a programming language. However, such choice may express unintended opinions on the API.
+
+#### Guidelines
+
+- No `default` means that the value is mandatory
+- Primitive types are `string`, `int`, `bool` and `uint`
+- Complex pre-define types are `struct`, `option` and `array`
+- Primitive types are preferred to describe the API for simplicity, the implementator may prefer a native type (e.g. `string` vs `Multiaddr` object/struct)
```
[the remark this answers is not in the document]
fryorcraken: I am not convinced `Waku Message` needs to leak out of the API.

In terms of `Multiaddr`, I don't want to prescribe the API to have to only accept an object `Multiaddr` because in some language like JS, a `string` is often used and expected.

Hence not entirely sure the best way to express that. I gave it a go in [`adaba54` (#65)](https://github.com/waku-org/specs/pull/65/commits/adaba5428b3a100c0f6b05c4577ed915e1bbbb46)

standards/application/waku-api.md
```diff
@@ -0,0 +1,224 @@
...
+    description: "A Waku node instance."
+  
+  Config:
+    type: struct
+    fields:
+      mode: 
+        type: string
+        constraints: ["edge", "relay"]
+        description: "The mode of operation of the Waku node. Core protocols used by the node are inferred from this mode."
+      network_config:
```
fryorcraken, line 126: Should this be "ClusterConf" instead? as network conf may related to TCP/IP level conf?

standards/application/waku-api.md
```diff
@@ -0,0 +1,224 @@
...
+#### Guidelines
+
+- No `default` means that the value is mandatory
+- Primitive types are `string`, `int`, `bool`, `enum` and `uint`
+- Complex pre-defined types are:
+  - `struct`: object and other nested types
+  - `option`: a value that can be set or left null
+  - `array`: iterable object containing values of all the same type
+  - `multiaddr`: a libp2p multiaddress; may be an object or a string, most idiomatic approach depending on the language
+  - `result`: an enum type that either contain a return value (success), or an error (failure). The error is left to the implementor
```
weboko, line 65: we can specify somewhere that `result` is `recommended` way of communicating errors rather than `throw` but it can be left to implementations

standards/application/waku-api.md
```diff
@@ -0,0 +1,224 @@
...
+  - `error`: Left to the implementor on whether `error` types are `string` or `object` in the given language.
+- The first parameter of a function MAY be considered as the object on which a method is called. It is left to the implementor on whether an objective programming approach is idiomatic to the language.
```
weboko, line 67: this line seems to be too specific even though it is true for `nim` and `rust`

fryorcraken:
**Only comment node configuration and initialization** part of this PR.

The work here is highly controversial, so let's focus on one bit at a time. Discussions related to other functions such as `subscribe`, etc, should wait for their own PR.

standards/application/waku-api.md
fryorcraken: we can re-introduce it when we define function that needs a reference to some exiting Waku node instance.

standards/application/waku-api.md
```diff
@@ -0,0 +1,215 @@
...
+  WakuNode:
+    type: struct
+    description: "A Waku node instance."
+  
+  Config:
+    type: struct
+    fields:
+      mode: 
+        type: string
+        constraints: ["edge", "relay"]
```
fryorcraken, line 105: `core` or `standard` are good alternatives to `relay`.

standards/application/waku-api.md
```diff
@@ -0,0 +1,215 @@
...
+    type: AutoShardingConfig
+    fields:
+      numShardsInCluster: 8
+```
+
+#### Extended definitions
+
+If the `mode` set is `edge`, the initialised `WakuNode` MUST mount:
+
+- [LIGHTPUSH](https://github.com/vacp2p/rfc-index/blob/main/waku/standards/core/19/lightpush.md) as client
```
fryorcraken, line 182: Metadata should be loaded too.

standards/application/waku-api.md
```diff
@@ -0,0 +1,281 @@
...
+        constraints: [ "edge", "relay" ]
+        description: "The mode of operation of the Waku node. Core protocols used by the node are inferred from this mode."
+      network_config:
+        type: NetworkConfig
+        default: TheWakuNetworkPreset
+      store_confirmation:
+        type: bool
+        # Until further dogfooding, assuming default false, usage of SDS should be preferred
+        default: false
+        description: "No-payload store hash queries are made to confirm whether outbound messages where received by remote store node."
```
chaitanyaprem, line 115:
```suggestion
        description: "No-payload store hash queries are made to confirm whether outbound messages were received by remote store node."
```

standards/application/waku-api.md
```diff
@@ -0,0 +1,281 @@
...
+
+  NetworkConfig:
+    type: struct
+    fields:
+      boostrap_nodes:
+        type: array<string>
+        # Default means the node does not bootstrap, it is not ideal but practical for local development
+        # TODO: get feedback
+        default: ""
+        description: "Bootstrap nodes, entree and multiaddr formats are accepted."
```
chaitanyaprem, line 125: wondering if we should support specifying just enr's as well. 
I have noticed that statically configured peers via multiaddr alone sometimes get added to peer-exchange cache very late (this is because of underlying discv5 query taking a lot of time to get to random walk closer to this peer). Due to this peer discovery with capabilities takes time.

I had faced this issue while simulating mixnet with [chat app](https://github.com/waku-org/nwaku/blob/48097e935b75de483f2a1f51c7b1d659c8403b25/apps/chat2mix/config_chat2mix.nim#L85-L87) and had to statically pass ENR's so that capabilities also get added for statically discovered peers. Developers who want to run local setups would face similar issue and having this support would help them.

standards/application/waku-api.md
```diff
@@ -0,0 +1,281 @@
...
+      message_validation:
+        type: MessageValidation
+        # If the default config for TWN is not used, then we still provide a message validation default 
+        default: DefaultMessageValidation
+
+  AutoShardingConfig:
+    type: struct
+    fields:
+      num_shards_in_cluster:
+        type: uint
```
chaitanyaprem, line 149: should we make this uint16 as we know that is the max a shard number can be as per [spec](https://github.com/waku-org/specs/blob/master/standards/core/relay-sharding.md#static-sharding).

standards/application/waku-api.md
```diff
@@ -0,0 +1,281 @@
...
+
+  MessageValidation:
+    type: struct
+    fields:
+      max_message_size_bytes:
+        type: uint
+        default: 153600 # 150 KiB
+        description: "The maximum accepted message size in Bytes"
+      # For now, RLN is the only message validation available
+      rln_config:
```
chaitanyaprem, line 160: wondering if we should make it generic for users to specify validatorConfig and make RLN_Config a type of validator_config? 
or is it too early and we can do this at a later stage.

standards/application/waku-api.md
```diff
@@ -0,0 +1,281 @@
...
+- [STORE](https://github.com/vacp2p/rfc-index/blob/main/waku/standards/core/13/store.md) as client
+
+And must use mount and use the following protocols to discover peers:
+
+- [DISCV5](https://github.com/vacp2p/rfc-index/blob/main/waku/standards/core/33/discv5.md)
+- [PEER-EXCHANGE](https://github.com/vacp2p/rfc-index/blob/main/waku/standards/core/34/peer-exchange.md)
+- [RENDEZVOUS](https://github.com/waku-org/specs/blob/master/standards/core/rendezvous.md)
+
+`edge` mode SHOULD be used if node functions in resource restricted environment,
+whereas `relay` SHOULD be used if node has no strong hardware or bandwidth restrictions.
```
chaitanyaprem, line 263: maybe we can refer to roughly estimated bandwidth and hardware specified somewhere in specs?

standards/application/waku-api.md
chaitanyaprem: ok, saw your in progress comments from line-265 regarding this.

standards/application/waku-api.md
fryorcraken: > peer-exchange cache 

I see, you are saying that such a boostrap node would not be send to peer exchange clients, right?

chaitanyaprem: Yes, because its ENR is still not available.

standards/application/waku-api.md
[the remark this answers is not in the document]
fryorcraken: I think we cna expect it to be a `string` in most languages. so will do that.

standards/application/waku-api.md
fryorcraken: Yes, for now it's not generic. The expectation with RLN API deliverable is that the dev can pass an object that implement specific traits (`validate_message` etc) so it's more generic.

standards/application/waku-api.md
fryorcraken: I like the idea, but maybe best to do that later.

jm-clius (approved):
This LGTM now. Thanks!

standards/application/waku-api.md
```diff
@@ -0,0 +1,304 @@
...
+    fields:
+      remote_nodes:
+        type: array<string>
+        # Default means the node does not bootstrap, e.g. for local development
+        default: []
+        description: "Nodes to connect to; used for discovery bootstrapping and quick connectivity. entree and multiaddr formats are accepted."
+      static_store_nodes:
+        type: array<string>
+        default: []
+        description: "Only the passed nodes are used for store queries, discovered store nodes are discarded."
```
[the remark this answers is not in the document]
fryorcraken: It is really **only** about store nodes. See Status' current usage of store nodes. Aka, coping with the fact that store is not decentralised enough.

> "List of relay nodes to exchange information with. Other discovered nodes are discarded."

Not sure I understand this proposal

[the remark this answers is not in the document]
fryorcraken: But we don't want to have a more generic argument.

This is a temporary work around the fact that store decentralization and quality is not yet sorted.

I don't think we should offer a config where a user only connects to specific relay nodes (and basically disables all discovery).

standards/application/waku-api.md
```diff
@@ -0,0 +1,327 @@
...
+        default: "_platform dependent_"
+        description: "The mode of operation of the Waku node. Core protocols used by the node are inferred from this mode."
+      waku_config:
+        type: WakuConfig
+        default: TheWakuNetworkPreset
+      message_confirmation:
+        type: array<string>
+        constraints: [ "store", "filter" ]
+        # Until further dogfooding, assuming default false, usage of SDS should be preferred
+        default: [ "none" ]
```
Ivansete-status, line 132: Maybe :)?
```suggestion
        constraints: [ "none", "store", "sds" ]
        # Until further dogfooding, assuming default none, usage of SDS should be preferred
        default: [ "none" ]
```
Besides, I can't quite get why an `array<string>` is needed

standards/application/waku-api.md
```diff
@@ -0,0 +1,304 @@
...
+        type: NetworkConfig
+        default: DefaultNetworkingConfig
+      eth_rpc_endpoints:
+        type: array<string>
+        description: "Eth/Web3 RPC endpoint URLs"
+
+  WakuConfig:
+    type: struct
+    fields:
+      remote_nodes:
```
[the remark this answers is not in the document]
Ivansete-status: I like `entry_nodes` term as well :D

standards/application/waku-api.md
```diff
@@ -0,0 +1,327 @@
...
+Hence, instead of having the reader learn a new IDL, we propose to use a simple IDL with self-describing syntax.
+
+An alternative would be to choose a programming language. However, such choice may express unintended opinions on the API.
+
+### Primitive types and general guidelines
+
+- No `default` means that the value is mandatory, meaning a `default` value implies an optional parameter.
+- Primitive types are `string`, `int`, `bool`, `enum` and `uint`
+- Complex pre-defined types are:
+  - `struct`: object and other nested types.
```
weboko, line 71: nit: by being biased by high level languages I would prefer `object` instead of `struct`

standards/application/waku-api.md
```diff
@@ -0,0 +1,327 @@
...
+        constraints: [ "edge", "relay" ]
+        default: "_platform dependent_"
+        description: "The mode of operation of the Waku node. Core protocols used by the node are inferred from this mode."
+      waku_config:
+        type: WakuConfig
+        default: TheWakuNetworkPreset
+      message_confirmation:
+        type: array<string>
+        constraints: [ "store", "filter" ]
+        # Until further dogfooding, assuming default false, usage of SDS should be preferred
```
weboko, line 131: I don't understand why we mention `SDS` here as it is in `dogfooding` / `experimental` stage too

standards/application/waku-api.md
weboko: > Besides, I can't quite get why an array<string> is needed

@Ivansete-status as I understand it is because we can have multiple confirmation mechanisms working at the same time `filter` + `store`

standards/application/waku-api.md
```diff
@@ -0,0 +1,327 @@
...
+        description: "Eth/Web3 RPC endpoint URLs"
+
+  WakuConfig:
+    type: struct
+    fields:
+      seed_nodes:
+        type: array<string>
+        # Default means the node does not bootstrap, e.g. for local development
+        default: []
+        description: "Nodes to connect to; used for discovery bootstrapping and quick connectivity. entree and multiaddr formats are accepted."
```
weboko, line 149: nit: and we can remove the comment above

```suggestion
        description: "Nodes to connect to; used for discovery bootstrapping and quick connectivity. entree and multiaddr formats are accepted. If not provided, node does not bootstrap (e.g for local development)"
```

standards/application/waku-api.md
```diff
@@ -0,0 +1,327 @@
...
+      static_store_nodes:
+        type: array<string>
+        default: []
+        description: "Only the passed nodes are used for store queries, discovered store nodes are discarded."
```
weboko, line 153: sorry if you mentioned it somewhere and I didn't see

question: why discarded? previously we were discussing it in the way that discovered nodes will be de-prioritized and used only if static store nodes are not available

standards/application/waku-api.md
```diff
@@ -0,0 +1,327 @@
...
+      cluster_id:
+        type: uint
+      auto_sharding_config:
+        type: AutoShardingConfig
+        default: DefaultAutoShardingConfig
+        description: "The auto-sharding config, if sharding mode is `auto`"
+      message_validation:
+        type: MessageValidation
+        # If the default config for TWN is not used, then we still provide a message validation default
```
weboko, line 162:
```suggestion
        description: If the default config for TWN is not used, then we still provide a message validation default 
```

standards/application/waku-api.md
```diff
@@ -0,0 +1,327 @@
...
+    type: NetworkConfig
+    fields:
+      listen_address: "0.0.0.0"
+      p2p_tcp_port: 60000
+      discv5_udp_port: 9000
+
+  TheWakuNetworkPreset:
+    type: WakuConfig
+    fields:
+      remote_nodes: [ "enrtree://AIRVQ5DDA4FFWLRBCHJWUWOO6X6S4ZTZ5B667LQ6AJU6PEYDLRD5O@sandbox.waku.nodes.status.im" ]
```
weboko, line 247: I am not sure it worth mentioning `remote nodes` and `contract address` as they can get outdated rather fast: unpredicted contract changes like we did recently etc.

standards/application/waku-api.md
fryorcraken: I think we can re-adjust at implementation time. Will correct and add todo.

PR #66, opened 2025-06-19
chore: some suggestions for waku api

Ivansete-status:
This just serves as a simple way to share ideas about `waku-api`

### DIFF
--- a/standards/application/waku-api.md
+++ b/standards/application/waku-api.md
@@ -56,7 +56,7 @@ An alternative would be to choose a programming language. However, such choice m
 #### Guidelines
 
 - No `default` means that the value is mandatory
-- Primitive types are `string`, `int`, `bool` and `uint`
+- Primitive types are `string`, `int`, `bool`, `uint`, and `pointer`
 - Complex pre-define types are `struct`, `option` and `array`
 - Primitive types are preferred to describe the API for simplicity, the implementator may prefer a native type (e.g. `string` vs `Multiaddr` object/struct)
 
@@ -107,7 +107,6 @@ language_mappings:
     module_type: "esm"
 ```
 
-
 ### Type definitions
 
 ```yaml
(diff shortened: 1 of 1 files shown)


standards/application/waku-api.md
```diff
@@ -107,90 +107,152 @@ language_mappings:
...
+                       - waku goes online/offline.
+                       - waku is not responding for too long.
     returns:
-        type: result<void, string>
+        type: pointer
+        description: On success, returns a reference to the Waku context, that will be needed in
+                     the subsequent Waku API calls.
+                     On error, returns a NULL, and the detail will be given in the `init_callback`.
+
+  subscribe:
```
Ivansete-status, line 201: I think `subscribe` and `unsubscribe` should better belong to `messaging-api`
I left them just to share the format I would expect.

standards/application/waku-api.md
```diff
@@ -56,7 +56,7 @@ An alternative would be to choose a programming language. However, such choice m
 #### Guidelines
 
 - No `default` means that the value is mandatory
-- Primitive types are `string`, `int`, `bool` and `uint`
+- Primitive types are `string`, `int`, `bool`, `uint`, and `pointer`
```
fryorcraken, line 59: I would disagree with `pointer` because this is now very implementation dependent and idiomatic to the language. 

Also let's review if we really need it.

standards/application/waku-api.md
```diff
@@ -107,90 +107,152 @@ language_mappings:
...
-        description: "The mode of operation of the Waku node. Core protocols used by the node are inferred from this mode."
+        constraints: ["client", "server"]
+        description: "The mode of operation of the Waku node.
+                      Waku core protocols used by the node are inferred from this mode."
       network_config:
         type: struct
-        default: TheWakuNetworkPreset
         fields:
-          boostrap_nodes:
+          boostrap_servers:
```
fryorcraken, line 125: why? we never call this a "server"

standards/application/waku-api.md
```diff
@@ -107,90 +107,152 @@ language_mappings:
...
 types:
   Config:
     type: struct
     fields:
-      operating_mode: 
+      mode:
         type: string
-        constraints: ["edge", "relay"]
-        description: "The mode of operation of the Waku node. Core protocols used by the node are inferred from this mode."
+        constraints: ["client", "server"]
```
fryorcraken, line 119: how is a relay node a "server" node? We had extensive discussions around the topic here.

- client: what do you mean by "client"? the "light client" terminology barely apply to Waku, which is why after several discussions we settled on "edge". because a relay node may also be a "client" for store and peer exchang eservices
- server: how is a node that runs relay, is a "server"?

Note that we are **not** including "Waku as a service node" in this present API.

standards/application/waku-api.md
```diff
@@ -107,90 +107,152 @@ language_mappings:
...
         type: struct
-        default: TheWakuNetworkPreset
         fields:
-          boostrap_nodes:
+          boostrap_servers:
             type: array<string>
-            default: ""
-            description: "Bootstrap nodes, entree and multiaddr formats are accepted."
-          static_store_nodes:
+            default: "enrtree://AIRVQ5DDA4FFWLRBCHJWUWOO6X6S4ZTZ5B667LQ6AJU6PEYDLRD5O@sandbox.waku.nodes.status.im"
```
fryorcraken, line 127: I disagree with this approach because these are not independent default values, they are default values for TWN.

Hence, I propose to continue with the original approach to define a `TheWakuNetwork` hardcoded value`. See original PR.

standards/application/waku-api.md
```diff
@@ -107,90 +107,152 @@ language_mappings:
...
             type: array<string>
             default: []
-            description: "Only the passed nodes are used for store queries, discovered store nodes are discarded."
+            description: "Array of server nodes' multiaddresses. Discovered nodes may also be considered."
+            examples:
+              ex1: ["/dns4/node-01.do-ams3.waku.sandbox.status.im/tcp/30303/p2p/16Uiu2HAmNaeL4p3WEYzC9mgXBmBWSgWjPHRvatZTXnp8Jgv3iKsb"]
+              ex2: ["/dns4/node-01.do-ams3.waku.sandbox.status.im/tcp/8000/wss/p2p/16Uiu2HAmNaeL4p3WEYzC9mgXBmBWSgWjPHRvatZTXnp8Jgv3iKsb"]
           clusterId:
             type: uint
+            constraints: mode == "server"
```
fryorcraken, line 142: All Waku node need to have a cluster id set, why is this constraint here?

standards/application/waku-api.md
```diff
@@ -107,90 +107,152 @@ language_mappings:
...
             default: 1
-          sharding_mode:
-            constraints: ["auto", "static"]
-          auto_sharding_config:
-            type: optionAutoShardingConfig
-            default: none
-            description: "The auto-sharding config, if sharding mode is `auto`"
-          active_relay_shards:
+            description: "The cluster subscribed to and participating in. A cluster is composed by multiple shards."
+          shards:
```
fryorcraken, line 145: You seem to have ignored the original PR. The concept of `shards` in a configuration is very ambiguous. Are we talkin about shards available in the cluster? are we talking about shards for relay to subscribe to?

standards/application/waku-api.md
```diff
@@ -107,90 +107,152 @@ language_mappings:
...
+    type: uint
+    constraints: [0 (Ok), 1 (Error), 2 (MissingCallback), 3 (WakuNotResponding)]
+    description:
+      - 0: The function call was successful and finished on time.
+      - 1: Generic error when invoking a particular function. The error detail will be given in the msg callback field.
+      - 2: If, when invoking a certain function, a callback is not set properly.
+      - 3: When the waku node is blocked in a certain callback, or due to other reason, for too long.
+
+  WakuCallback:
+    type: pointer
```
fryorcraken, line 160: This is a function before being a pointer.

standards/application/waku-api.md
```diff
@@ -107,90 +107,152 @@ language_mappings:
...
 ```yaml
 functions:
   init:
-    description: "Initialise the waku node"
+    description: "Initialise the waku instance"
     parameters:
         - name: config
           type: Config
-          description: "The Waku node configuration."
+          description: "The Waku configuration."
```
fryorcraken, line 185: I am not sure to understand the subtle changes you are proposing.

standards/application/waku-api.md
```diff
@@ -107,90 +107,152 @@ language_mappings:
...
+            default: [0, 1, 2, 3, 4, 5, 6, 7]
+            description: "The shards subscribed to and participating in."
+  ErrorCode:
+    type: uint
+    constraints: [0 (Ok), 1 (Error), 2 (MissingCallback), 3 (WakuNotResponding)]
+    description:
+      - 0: The function call was successful and finished on time.
+      - 1: Generic error when invoking a particular function. The error detail will be given in the msg callback field.
+      - 2: If, when invoking a certain function, a callback is not set properly.
+      - 3: When the waku node is blocked in a certain callback, or due to other reason, for too long.
```
fryorcraken, line 157: I suggest that for now, errors, their types and values are left to the implementor

standards/application/waku-api.md
Ivansete-status: > I am not sure to understand the subtle changes you are proposing.

Yes, from `nwaku` PoV, we are talking about Waku instances, not waku nodes.

Waku is composed by:
- waku node
- health monitor
- discovery service

In turn, waku node is just a libp2p switch and a set of libp2p protocols.

In the waku api we are exposing Waku instances and not just waku nodes. Hence, I think is more accurate to only use the waku term.

standards/application/waku-api.md
```diff
@@ -107,90 +107,152 @@ language_mappings:
...
-    description: "Initialise the waku node"
+    description: "Initialise the waku instance"
     parameters:
         - name: config
           type: Config
-          description: "The Waku node configuration."
+          description: "The Waku configuration."
+        - name: callback
+          type: WakuCallback
+          description: Gives feedback in case of failure when invoking the function.
```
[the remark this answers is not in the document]
fryorcraken: Can you please expand? a `result` pattern is simpler and preferred when doing a call that can return ONE error.

The object you can return in a callback or in a result is the same. how is it richer?

standards/application/waku-api.md
Ivansete-status: The `pointer` type is needed because it is the type returned by `waku_new`, which return a raw memory reference to the created Waku context.

standards/application/waku-api.md
Ivansete-status: Every function is a pointer, isn't it :) ?

standards/application/waku-api.md
Ivansete-status: Yup, you are right !

standards/application/waku-api.md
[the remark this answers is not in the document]
Ivansete-status: The main reason why we use callback to return richer feedback is to avoid having memory leaks, i.e., every memory allocated within the libwaku should be deallocated within libwaku itself. We cannot have library functions returning complex structures. It just return `int`.

```nim
// The possible returned values for the functions that return int
#define RET_OK                0
#define RET_ERR               1
#define RET_MISSING_CALLBACK  2
```

Notice though, that this suggestion considers a very low C level.

Therefore, I think you are absolutely right about the `Result` pattern and I believe we can apply it at higher API levels.

In fact, we are applying it in `waku-rust-bindings`. See for example: https://github.com/waku-org/waku-rust-bindings/blob/master/waku-bindings/src/node/management.rs

standards/application/waku-api.md
```diff
@@ -107,90 +107,152 @@ language_mappings:
...
+          static_servers:
             type: array<string>
             default: []
-            description: "Only the passed nodes are used for store queries, discovered store nodes are discarded."
+            description: "Array of server nodes' multiaddresses. Discovered nodes may also be considered."
+            examples:
+              ex1: ["/dns4/node-01.do-ams3.waku.sandbox.status.im/tcp/30303/p2p/16Uiu2HAmNaeL4p3WEYzC9mgXBmBWSgWjPHRvatZTXnp8Jgv3iKsb"]
+              ex2: ["/dns4/node-01.do-ams3.waku.sandbox.status.im/tcp/8000/wss/p2p/16Uiu2HAmNaeL4p3WEYzC9mgXBmBWSgWjPHRvatZTXnp8Jgv3iKsb"]
```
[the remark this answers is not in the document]
Ivansete-status: The bootstrap node represents is a super thin node which doesn't perform any Waku protocol action. It is just a discovery facilitator.

In turn, a static_server (or static_node), is a fixed well-known node that can participate on any Waku protocol (`relay`, `store`, etc.)

standards/application/waku-api.md
jm-clius: > I think @jm-clius can shed more light as to why we consider that: waku node is just a libp2p switch and a set of libp2p protocols

This is just the internal view to properly separate concerns and initialisation. The `WakuNode` object (in Nim) wraps a libp2p `switch` and mounts various libp2p protocols onto this switch. Composing the various discovery models, startup/config logic, etc. into this `object` caused many circular references and complex code. In other words, it's an implementation detail. However, externally the compiled library and API is still seen as a single, cohesive thing - a "Waku Node" in other words.

standards/application/waku-api.md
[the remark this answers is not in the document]
jm-clius: Oh! No servers in Waku, please. 😝  Where we treat nodes as server-like, it's a temporary error.

standards/application/waku-api.md
fryorcraken: After spending more time in the code, I would suggest that from an API PoV, A "Waku Node" is:

- libp2p switch and related protocols implementations
- discv5 service (other discovery services are libp2p protocols)
- health monitor
- metrics

Then, as per your previous comment around the REST API @Ivansete-status, I would say that the REST API service is **not** part of a Waku node, and instead should be seen as a external library that one may want to use with a Waku node implementation.

Which makes it very interesting because in terms of clean up for the REST API (which we have **not** scheduled yet), it means that will be force to only expose on the REST API, what is exposed by nwaku, and the REST API will not be able to access "internal", allowing for a better design. cc @NagyZoltanPeter

PR #74, opened 2025-07-22
Help implementer by making the implementation section less optional

fryorcraken:


fryorcraken:
> Am I right that the intention here is indeed just to make it "less optional" as the description says? Will be willing to approve if this is not meant to make this section mandatory, but just to encourage more editors to make use of it. I agree that adding more implementation detail could be useful, but I'm not convinced that this section makes sense for every type of spec (especially if we want to lower barrier to writing more raw specs, that doesn't even require an implementation).

I think here I need to leave the call to you as you're overseeing most spec writing.

jm-clius (approved):

Issue #75, opened 2025-07-24
Extend sharding specification


Ivansete-status:
The current [_Relay Sharding_ specification](https://github.com/waku-org/specs/blob/9992da00d8b31c75f20d2c9a5762929606ebf2dc/standards/core/relay-sharding.md) provides a solid definition of the various sharding mechanisms.

However, there is some room for improvement and additional clarification to ensure a consistent approach across all Waku implementations.

Key points for consideration include:
 
- Deprecation of "Named Sharding": Explicitly mark the "Named Sharding" concept as deprecated, as it it no longer in use.
 
- Default sharding mechanism: While _auto sharding_ is described as the default mechanism, this primarily seems to apply to "network consumers", i.e., members that publish or receive messages. For node operators, i.e., members responsible for relaying and validating messages, _static sharding_ seems to be the best default.

- num-shards-in-network definition: There is no clear guidance on the appropriate `num-shards-in-network` value to use and any given time, or whether the maximum value (`1024`) should be the default.

- Scaling strategy: Provide a clear explanation of the recommended strategy for scaling networks up and down.

- PubsubTopic visibility: We need to clarify whether the pubsubTopic concept should be exposed to Waku consumers or reserved for internal implementations only. In the latter case, Waku integrators should work exclusively with the cluster-id and shard terms.

In summary, the current specification does an excellent job of describing the individual sharding mechanisms, but a clear definition of how to properly combine those ingredients is still very much needed :D

PR #76, opened 2025-07-24
docs: some clarification on defaults for sharding

jm-clius:
Closes https://github.com/waku-org/specs/issues/75

Clarified that:
- named sharding is deprecated
- default sharding is _static_ sharding
- autosharding only makes sense for defined networks. The default network should be TWN. All other networks should be assumed to have only `1` shard by default.
- pubsub topics should not be visible to app-level protocols

### DIFF
--- a/standards/core/relay-sharding.md
+++ b/standards/core/relay-sharding.md
@@ -223,9 +223,21 @@ Generation default value is `0`.
 - Full length `/0/myapp/1/mytopic/cbor`
 - Short length `/myapp/1/mytopic/cbor`
 
-#### Generation
+#### Generation and scaling
 
-The generation number monotonously increases and indirectly refers to the total number of shards of the Waku Network.
+The generation number monotonously increases and indirectly refers to the total number of shards of a defined network.
+In order to scale,
+each subsequent generation of a defined network can define a larger `number_of_shards_in_network`,
+with the content topics only sharded to the number of shards defined for the corresponding generation of the network.
+Generational autosharding MUST be clearly defined for each generation of a network.
+
+For example, consider a specific network defined as the 2 shards `0` and `1` in cluster `32`.
+This will automatically be assumed to be generation `0` of this network.
+All content topics in the format `/myapp/1/mytopic/cbor` defined for apps on this network will be autosharded to 1 of these 2 shards.
+If in future the specifiers of this network want to scale the network to 4 shards (i.e. shards `0` to `3` on cluster `32`),
+they MUST define a generation `1` version of the network with `number_of_shards_in_network = 4`.
+New content topics for apps on generation `1` of this network MUST be prefixed in the format `/1/myapp/1/mytopic/cbor` to be autosharded into all 4 shards.
+Legacy generation `0` content topics will still only be autosharded into the original 2 shards.
 
 <!-- Create a new RFC for each generation spec. -->
 
--- a/standards/core/relay-sharding.md
+++ b/standards/core/relay-sharding.md
@@ -34,7 +34,7 @@ This document also covers discovery of topic shards.
 
 ## Named Sharding
 
-> **_Note:_** As of 2025-07-24 named sharding has been deprecated from all Waku implementations.
+> **_Note:_** As of 2025-07-24 freely named sharding has been deprecated from all Waku implementations.
 It is still described here as background for static and automatic sharding.
 
 _Named sharding_ offers apps to freely choose pubsub topic names.
@@ -52,14 +52,15 @@ _Static sharding_ is an extension of named sharding that offers a set of shards 
 Assigning content topics to specific shards is up to app protocols,
 but the discovery of these shards is managed by Waku.
 This is the RECOMMENDED default format for shards chosen by an app protocol.
-It is RECOMMENDED that, for simplification of configuration and various APIs,
-all app-level protocols only interact with `cluster` and `shard`
-and never the fully-formed pubsub topic,
-which is a concern internal to the Waku core.
 
 Static shards are managed in shard clusters of 1024 shards per cluster.
 Waku static sharding can manage $2^{16}$ shard clusters.
 Each shard cluster is identified by its index (between $0$ and $2^{16}-1$).
+
+It is RECOMMENDED that, for simplification of configuration and various APIs,
+all app-level protocols only interact with `cluster` and `shard`
+and never the fully-formed pubsub topic,
+which is a concern internal to the Waku implementation.
 
 A specific shard cluster is either globally available to all apps,
 specific for an app protocol,
@@ -84,7 +85,7 @@ but choosing shards in the global cluster has a higher probability of sharing th
 This offers k-anonymity and better connectivity, but comes at a higher bandwidth cost.
 
 Since the introduction of [the Waku Network](https://github.com/vacp2p/rfc-index/blob/main/waku/standards/core/64/network.md),
-it is RECOMMENDED that apps choose a shard within the defined range for that network.
+it is RECOMMENDED that apps choose a cluster + shard within the defined range for that network.
 
 The name of the pubsub topic corresponding to a given static shard is specified as
 
--- a/standards/core/relay-sharding.md
+++ b/standards/core/relay-sharding.md
@@ -34,9 +34,13 @@ This document also covers discovery of topic shards.
 
 ## Named Sharding
 
+> **_Note:_** As of 2025-07-24 named sharding has been deprecated from all Waku implementations.
+It is still described here as background for static and automatic sharding.
+
 _Named sharding_ offers apps to freely choose pubsub topic names.
 It is RECOMMENDED for App protocols to follow the naming structure detailed in [23/WAKU2-TOPICS](https://github.com/vacp2p/rfc-index/blob/main/waku/informational/23/topics.md).
 With named sharding, managing discovery falls into the responsibility of apps.
+For this reason it is NOT RECOMMENDED that app protocols and Waku implementations make use of named sharding.
 
 From an app protocol point of view, a subscription to a content topic `waku2/xxx` on a shard named /mesh/v1.1.1/xxx would look like:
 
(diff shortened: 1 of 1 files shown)


standards/core/relay-sharding.md
```diff
@@ -166,9 +175,19 @@ This example node is part of shards `13`, `14`, and `45` in the Status main-net
 
 ## Automatic Sharding
 
-Autosharding selects shards automatically and is the default behavior for shard choice.
-The other choices being static and named sharding as seen in previous sections.
-Shards (pubsub topics) SHOULD be computed from content topics with the procedure below.
+Autosharding is an extension of static sharding, 
+where shards can automatically be selected based on content topic.
+This is only possible in cases where there exists a clear definition of a "network" combining multiple shards
+with a predefined cluster ID and number of shards in the network (`number_of_shards_in_network`).
```
Ivansete-status, line 181: This is a very good sentence!
I'm adding a suggestion for more guidance about what is the preferred sharding mechanism depending on the node's type.
```suggestion
This is only possible in cases where there exists a clear definition of a "network" combining multiple shards
with a predefined cluster ID and number of shards in the network (`number_of_shards_in_network`).
Therefore, autosharding is only reserved for edge clients who know the network topology, i.e., `number_of_shards_in_network`, whereas, it is assumed that relay nodes are running using static sharding.
```

standards/core/relay-sharding.md
```diff
@@ -204,9 +223,21 @@ Generation default value is `0`.
...
+
+The generation number monotonously increases and indirectly refers to the total number of shards of a defined network.
+In order to scale,
+each subsequent generation of a defined network can define a larger `number_of_shards_in_network`,
+with the content topics only sharded to the number of shards defined for the corresponding generation of the network.
+Generational autosharding MUST be clearly defined for each generation of a network.
 
-The generation number monotonously increases and indirectly refers to the total number of shards of the Waku Network.
+For example, consider a specific network defined as the 2 shards `0` and `1` in cluster `32`.
+This will automatically be assumed to be generation `0` of this network.
```
Ivansete-status, line 235: I've got the impression that we need to explicitly set the `generation` field as part of the pubsub-topic definition.
F.e., `/waku/<generation>/<cluster-id>/<shard-number>`, and maybe part of the node's network config.

The reason for that is that there might be mismatches among nodes running in different generation.

And, regarding the `generation` concept:
1. What is the measure to consider in order to take a decision of switching from one generation to another
2. What happens if there are nodes running from different generations and how to enforce all nodes use the latest generation possible?
3. How are we effectively and mathematically solving the scaling problem by incrementing the generation number and the `number_of_shards_in_network`?

standards/core/relay-sharding.md
```diff
@@ -204,9 +223,21 @@ Generation default value is `0`.
...
+All content topics in the format `/myapp/1/mytopic/cbor` defined for apps on this network will be autosharded to 1 of these 2 shards.
+If in future the specifiers of this network want to scale the network to 4 shards (i.e. shards `0` to `3` on cluster `32`),
+they MUST define a generation `1` version of the network with `number_of_shards_in_network = 4`.
```
Ivansete-status, line 238: Ah okay, nevertheless, I think is better to have the `generation` field as part of the resulting pubsub-topic because the `generation` concept applies to both `static` and `auto` sharding

Ivansete-status (commented):
Thanks so much for it! :raised_hands: 
Adding some comments and questions :D

standards/core/relay-sharding.md
jm-clius: Ah, I think there might still be some confusion as to what the `generation` concept implies. It's just a convenient way for us to show that we _could_ expand a network definition in future to include more shards than it does currently (usually to allow for scaling). The shards of the previous generation(s) are not affected (and should not have different names). This concept is only really important for autosharding as a new generation allows the content topics to map to more shards than they do currently, without affecting backwards compatibility. In other words, a content topic in one generation will not suddenly map to a different shard once we have a new generation of shards in the definition. Outside of autosharding, the "generation" just suggests relay and service nodes which shards to subscribe to to support the network. Remember though that there's no issue with relay nodes subscribing to fewer or more shards. Gossipsub/Relay deals well with a heterogenous environment where each node selects which shards they subscribe to. I'm sure this may still be a bit obscure, so happy to explain in more detail if you have more questions.

In terms of these questions:
1. There is no definition (and shouldn't be, because that would be jumping the gun). A specific network, say TWN, will have to come up with their own metric on when to roll out a new generation. This is likely when the bandwidth required to relay one of the network shards become unacceptably high (~10Mbps). This is not an automatic process but a function of a network specification and consensus.
2. Nodes do not run within a generation - it's only the autosharding API for clients that has this logic embedded. Presumably if a client app wants to have their application be routed on one of the new generation's shards it will use the latest Waku node that supports this generation of autosharding (otherwise I think the API will error out). Relay nodes do _subscribe_ to shards that belong to one of the networks' generations, but remember that gossipsub already make sure that nodes that subscribe to the same shards belong to the same mesh. We can support a mixed topology where relay and service nodes support multiple shards (from multiple generations). In other words, we don't have to enforce anything - just encourage nodes to upgrade to have more nodes supporting the new shards.
3. We're not necessarily doing it most effectively and our mathematics are purposefully naive. If you use autosharding, a new content topic has equal chance of being mapped to any of the shards currently defined for the network. If there are 8 shards we believe that roughly each shard will carry 1/8th of the network traffic. If that becomes too much, we roll out, say, 24 new shards, and now each _new_ content topic has an equal chance of being mapped to any of the 8 + 24 = 32 shards. Therefore shards now carry 1/32 of traffic. A node has to support a minimum of 1 shard and can therefore participate with much lower bandwidth requirement.

standards/core/relay-sharding.md
```diff
@@ -34,19 +34,28 @@ This document also covers discovery of topic shards.
...
+For this reason it is NOT RECOMMENDED that app protocols and Waku implementations make use of named sharding.
 
 From an app protocol point of view, a subscription to a content topic `waku2/xxx` on a shard named /mesh/v1.1.1/xxx would look like:
 
 `subscribe("/waku2/xxx", "/mesh/v1.1.1/xxx")`
 
 ## Static Sharding
 
-_Static sharding_ offers a set of shards with fixed names.
+_Static sharding_ is an extension of named sharding that offers a set of shards with _fixed_ names.
```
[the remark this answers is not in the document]
weboko: I mean sentence on line `43` and not on `38`

standards/core/relay-sharding.md
```diff
@@ -34,19 +34,28 @@ This document also covers discovery of topic shards.
...
 Assigning content topics to specific shards is up to app protocols,
 but the discovery of these shards is managed by Waku.
+This is the RECOMMENDED default format for shards chosen by an app protocol.
+It is RECOMMENDED that, for simplification of configuration and various APIs,
+all app-level protocols only interact with `cluster` and `shard`
+and never the fully-formed pubsub topic,
+which is a concern internal to the Waku core.
```
weboko, line 58: maybe `libraries` instead of `core`, I believe word `core` might be misunderstood by people who didn't participate in our discussions around APIs etc.

```suggestion
which is a concern internal to the Waku libraries.
```

standards/core/relay-sharding.md
```diff
@@ -166,9 +175,19 @@ This example node is part of shards `13`, `14`, and `45` in the Status main-net
...
+Autosharding is an extension of static sharding, 
+where shards can automatically be selected based on content topic.
+This is only possible in cases where there exists a clear definition of a "network" combining multiple shards
+with a predefined cluster ID and number of shards in the network (`number_of_shards_in_network`).
+An example is [the Waku Network](https://github.com/vacp2p/rfc-index/blob/main/waku/standards/core/64/network.md).
+To use autosharding,
+the shards corresponding to the network definition MUST be numbered from `0` and monotonously increase to `number_of_shards_in_network - 1`.
+The Waku Network SHOULD be considered the default destination for all app-level protocols.
+In cases where the Waku Network is not used but autosharding is desired,
+the default number of shards in a network SHOULD be considered `1`.
```
weboko, line 187: I'd like to understand this sentence. 
Why would someone need `autosharding` with number of shard to be `1`?

standards/core/relay-sharding.md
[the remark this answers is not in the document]
fryorcraken: I think it's fine.

standards/core/relay-sharding.md
fryorcraken: `implementations` might be most accurate.

standards/core/relay-sharding.md
```diff
@@ -74,6 +83,9 @@ The global shard with index 0 and the "all app protocols" range are treated in t
 but choosing shards in the global cluster has a higher probability of sharing the shard with other apps.
 This offers k-anonymity and better connectivity, but comes at a higher bandwidth cost.
 
+Since the introduction of [the Waku Network](https://github.com/vacp2p/rfc-index/blob/main/waku/standards/core/64/network.md),
+it is RECOMMENDED that apps choose a shard within the defined range for that network.
```
fryorcraken, line 87: Hum, are we saying we assume they are running on the TWN? Do you mean choose a cluster and shard?

standards/core/relay-sharding.md
fryorcraken: Strongly disagree here. Relay nodes should also be using auto-sharding.

I guess we did not align on the previous debate: From an app developer point of view, I integrate Waku in my desktop app. I set my content topics for my app. I expect the relay node running in my desktop app to subscribe to shards based on content topics.

A relay node is not running `static sharding`, the API should still be content topic based.

standards/core/relay-sharding.md
fryorcraken: Because it allows usage of `content topic` only APIs, where `shard id` is never exposed or used by the developer.

fryorcraken (approved):
Excellent.

jm-clius:
Thanks everyone for your comments. I think I've addressed everything that was blocking. Please feel free to raise any further queries with me and we can address in a follow-up.

standards/core/relay-sharding.md
Ivansete-status: Thanks so much for the comments @jm-clius :raised_hands: !

---------

> The shards of the previous generation(s) are not affected (and should not have different names). This concept is only really important for autosharding as a new generation allows the content topics to map to more shards than they do currently, without affecting backwards compatibility

The above statement is pending to be implemented. Shall we create a deliverable to cover that?
https://github.com/waku-org/nwaku/blob/dbecd721f9f86e8d05f1506126033fcfa206c084/waku/waku_core/topics/sharding.nim#L32-L43

Furthermore, it seems there should be a tight link between the generation concept and the "num shards in cluster" but this is not the case nowadays, i.e., the "num shards in cluster" can be freely changed in the nwaku library but still considering the zero generation (see the whole sharding.nim I shared above.)

I've got the impression that we need to iterate further on that topic ;P or at least, properly reflect in code what we have from research PoV, and I'm more than happy to help on that matter

standards/core/relay-sharding.md
Ivansete-status: Thanks @jm-clius and makes sense 100%!
Let's imagine the network has evolved to Gen 44 , with 100.000 topics. Will a node be able to properly handle that load effectively? ( even 100 topics sound a lot .)

That's something to consider in the future but I am not still convinced we are solving the scaling problem yet but I might be missing something :D

PR #77, opened 2025-07-25
New RFC: WAKU-APP-QAKU

jimstir:
The Waku web app RFC for Qaku (questions and answers over Waku)....


danisharora099 (commented):
Overall comment at this stage is that it would be helpful to have definitions on what individual attributes on the types are

fbarbu15:
Hi @fryorcraken  @danisharora099   can we get another round of reviews please. I think all comments have been addressed
THanks

igor-sirotin:
As of today, Qaku is out of scope of Logos Messaging

PR #83, opened 2025-09-23
Chat Inbox

jazzz:
This PR adds a specification which describes a chat initialization protocol. 
 
 This replaces the previous effort #72 as the changes are substantial. This specification incorporates the feedback from the previous version - Specifically:
 - Removes Waku specific details (such as ContentTopic) and introduces a `delivery_address` abstraction
 - Removed `inbox_address` for simplicity
 - Defers `Conversation_id` semantics to a implementation specific specification. #TODO: Link
 - Defer envelope definition and framing strategy to implementation specific specification. #TODO: Link 
 - Adds clarity on MessageFlow
 
 In addition it also:
 - updates the encryption primitives to provide, mutual authentication.

### DIFF
--- a/standards/application/inbox.md
+++ b/standards/application/inbox.md
@@ -14,7 +14,7 @@ contributors:
 
 Communication protocols often face a fundamental bootstrapping problem: how can two parties establish secure, authenticated communication when they lack trusted channels? Traditional centralized approaches rely on servers to deliver invitations using pre-agreed upon encryption mechanisms. 
 
-In a decentralized context, this problem is more pronounced as participants cannot rely on an external service to help navigate version mismatches and client routing. Senders must know where participants are listening for messages and how to encode messages. Clients running older versions of the protocol are not able to interoperate with newer protocol versions which can fragment the network during upgrades.
+In a decentralized context, this problem is more pronounced as participants cannot rely on an external service to help navigate version mismatches and client routing. Senders must know where participants are listening for messages and how to encode them so they can be properly received. Clients running older versions of the protocol are not able to interoperate with newer protocol versions which can fragment the network during upgrades.
 
 
 # Theory / Semantics
@@ -33,7 +33,7 @@ The terms include:
 ## Inbox
 
 Inboxes are a mechanism for a client to receive messages from others which it does not share a channel with. 
-They are inbound only method to exchange payloads with a client. Its most direct application is to enable the reception of invites to other communication channels.
+They are an inbound only method to exchange payloads with a client. Its most direct application is to enable the reception of invites to other communication channels.
 
 They provide :
  - Payload Confidentiality: Only the participants can read the contents of any message sent.
--- a/standards/application/inbox.md
+++ b/standards/application/inbox.md
@@ -54,7 +54,7 @@ This protocol assumes that a delivery service is used to transport messages as d
 
 - Recipient subscribes to a `delivery_address`.
 - Recipient and Sender exchange the required data out of band.
-- Sender sends message to the `deliver_address` encrypted for the recipient.
+- Sender sends message to the `delivery_address` encrypted for the recipient.
 
 ```mermaid
 sequenceDiagram
@@ -87,9 +87,9 @@ Each participant has two keys:
 
 ## Pre-messages 
 
-As a prerequiste to establishing a secure channel, the `sender` must know the `recipients` InstallationKey and EphemeralKey
+As a prerequisite to establishing a secure channel, the `sender` must know the `recipients` InstallationKey and EphemeralKey
 
-How these are transfered is out of scope for this protocol.
+How these are transferred is out of scope for this protocol.
 
 
 ## Key Exchange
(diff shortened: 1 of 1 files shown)


standards/application/inbox.md
```diff
@@ -0,0 +1,190 @@
...
+    participant I as <delivery_address><br>(Raya)
+    actor R as Raya
+
+    I -->> R: Subscribe
+    
+    note over R,S: Send Out-of-Band
+    R ->> S : PreMessage(IK,EK,<delivery_address>)
+
+    note over R,S: Sent In-Band
+    S ->> I : M1
```
osmaczko: Is `M1` a `InboxHandshakeV1`?

standards/application/inbox.md
```diff
@@ -0,0 +1,190 @@
...
+    int64 timestamp = 1;
+    oneof frame_type {
+        invite.InvitePrivateV1 invite_private_v1 = 10;
+    }
+}
+```
+
+## Implementation Suggestions (optional)
+
+Recipients ephemeral key is not signed, and is susceptible to tampering by an active attacker during transit out of band. Implementors SHOULD keep the number of valid keys as little as possible. Revoking keys after a defined period of time, and only generating keys when needed minimizes the ability for an attacker to substitute weak keys
```
osmaczko: Is the unsigned EK risk here actually a confidentiality concern?Since the Recipient generates and tracks all their own EKs, a substituted EK would fail the key lookup on the Recipient side. So isn't this effectively a DoS (silently lost invite) rather than a "weak key" / forward secrecy issue?

Also, is the decision not to sign the ephemeral key still the case? According to https://github.com/logos-messaging/libchat/issues/27, the `IntroductionBundle` is expected to contain a signature.

osmaczko (commented):
The specification might benefit from explicitly referencing the Chat Framework, for example by:

- Stating that Inbox is an Initialization Protocol as defined by the Chat Framework
- Referring to the `PreMessage` as the `IntroductionBundle` for this protocol (or documenting the mapping)
- Clarifying that the out-of-band exchange of the `PreMessage` falls under the responsibility of the Discovery Protocol

standards/application/inbox.md
jazzz: `InboxHandshakeV1` is sufficient here. There is only one message type defined currently so yes `M1 = InboxHandshakeV1` but that is not categorically true into the future. 

This diagram is aimed at representing Message flow generically.

standards/application/inbox.md
jazzz: You are correct to point out this inconsistency. 

Originally This Spec used a XK0 Handshake which was removed in [this commit](https://github.com/logos-messaging/specs/pull/83/changes/1c4eb7d1ade9873350b3436a92e90cd33a5a3ead) in favor of X3DH.

This spec has not been fully updated.

standards/application/inbox.md
```diff
@@ -0,0 +1,190 @@
...
+
+[TODO]
+
+## Ciphersuite
+
+[TODO]
+
+## Recipient Key Identifier  
+When receiving a payload it is initially unclear which Recipients ephemeral key was used by the sender in the noise handshake. 
+The Recipient sends their ephemeral key out of band, and there may exist many such keys. The recipient could exhaustively attempt decryption with it's available keys however this is potentially inefficient. Additionally in the case of decryption failure, its ambiguious whether the ephemeral keys was never valid or has expired/consumed.
```
igor-sirotin:
```suggestion
The Recipient sends their ephemeral key out of band, and there may exist many such keys. The recipient could exhaustively attempt decryption with its available keys however this is potentially inefficient. Additionally in the case of decryption failure, its ambiguious whether the ephemeral keys was never valid or has expired/consumed.
```

standards/application/inbox.md
```diff
@@ -0,0 +1,190 @@
...
+
+To make this efficient, the Sender includes a short identifier for the key.
+
+Calculated as: `blake2b(utf8ToBytes('WAP') || K)`[0..4]
```
igor-sirotin, line 109: What is `WAP`? Waku Application Protocol? 🤔 
Not important at all, but would be nice to specify for clarity.

standards/application/inbox.md
```diff
@@ -0,0 +1,190 @@
...
+### Encryption Wrapper
+```protobuf
+message InboxHandshakeV1 {
+    message InboxHeaderV1 {
+        bytes initiator_static = 1;    
+        bytes initiator_ephemeral = 2;  
+        bytes responder_static = 3;    
+        bytes responder_ephemeral = 4; // Replace with RKI to save bytes
+    }
+    InboxHeader header = 1;
```
igor-sirotin, line 139:
```suggestion
    InboxHeaderV1 header = 1;
```

igor-sirotin (approved):
I see there's many TODOs, and the implementation went further and is now inconsistent with this spec, but I think we can get it merged and continue iterating on it later. It's still a raw spec anyway.

standards/application/inbox.md
jazzz: WAP => Waku Application Protocols ; It was the suggested product name before the Logos Rebranding. Its out of date and will need an update

PR #84, opened 2025-09-26
Update CI-spell to filter html

jazzz:
## Problem

The CI spellchecker flags html tags, and urls as "misspelled". This results in actions failing despite there being no spelling issues in the specification.

## Solution

- Adds a html filter 
- Sorts the word list to reduce conflicts in future PR's

Note: Word list is still a work in progress, adding more filters will reduce the number of entries required in the future


jm-clius (approved):
Thanks

PR #85, opened 2025-10-01
Rename Waku API's "Waku Config" to "Protocols" Config

fryorcraken:
Make it clearer that with this config, we are configuring the Waku protocols, in contrast to other parameters which are more executable related.

Related to https://github.com/waku-org/nwaku/issues/3493

Implementation: https://github.com/waku-org/nwaku/pull/3603


jm-clius (approved):
Sensible, thanks.

weboko (approved):

PR #91, opened 2025-10-19
feat: add segmentation spec

plopezlpz:
Added the specs for Application level segmentation of messages to be able to send messages larger than 150kb using Waku by partitioning them at source and reconstructing them on the recipient

Already implemented in https://github.com/waku-org/nim-chat-sdk/blob/main/chat_sdk/segmentation.nim


standards/application/segmentation.md
```diff
@@ -0,0 +1,181 @@
...
+title: Message Segmentation and Reconstruction over Waku
+name: Message Segmentation and Reconstruction
+tags: [waku-application, segmentation]
+version: 0.1
+status: draft
+---
+
+## Abstract
+
+This specification defines an application-layer protocol for **segmentation** and **reconstruction** of messages carried over Waku when the original payload exceeds the maximum Waku's message size. Applications partition the payload into multiple Waku envelopes and reconstruct the original on receipt, even when segments arrive out of order or up to **12.5%** of segments are lost. The protocol uses **Reed–Solomon** erasure coding for fault tolerance. Messages whose payload size is **≤ `segmentSize`** are sent unmodified.
```
kaichaosun: Not hard code here for 12.5%, use some term like `predefined rate`.

standards/application/segmentation.md
```diff
@@ -0,0 +1,181 @@
...
+
+## Motivation
+
+Waku Relay deployments typically propagate envelopes up to **1 MB**. To support larger application payloads (e.g., up to **10 MB** or more), a segmentation layer is required. This specification enables larger messages by partitioning them into multiple envelopes and reconstructing them at the receiver. Erasure-coded parity segments provide resilience against partial loss or reordering.
```
kaichaosun: The default seems to be 150kb by default in nwaku.
https://github.com/waku-org/nwaku/blob/9808e205af522989cc6888c12e6656df47e0da48/waku/waku_core/message/default_values.nim#L5

You may also want to update the message spec to make it clear and reference it here.

standards/application/segmentation.md
```diff
@@ -0,0 +1,181 @@
...
+- **Intended for:** all Waku nodes at the application layer
+
+---
+
+## Security Considerations
+
+### Privacy
+
+`entire_message_hash` enables correlation of segments that belong to the same original message but does not reveal content.  
+Applications **SHOULD** encrypt payloads before segmentation.
```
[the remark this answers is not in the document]
plopezlpz: removing so developers can do as they please, before or after or no encryption

standards/application/segmentation.md
```diff
@@ -0,0 +1,182 @@
...
+
+The key words **“MUST”**, **“MUST NOT”**, **“REQUIRED”**, **“SHALL”**, **“SHALL NOT”**, **“SHOULD”**, **“SHOULD NOT”**, **“RECOMMENDED”**, **“NOT RECOMMENDED”**, **“MAY”**, and **“OPTIONAL”** in this document are to be interpreted as described in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt).
+
+## Segmentation
+
+### Sending
+
+When the original payload exceeds `segmentSize`, the sender **MUST**:
+
+- Compute a 32-byte `entire_message_hash = Keccak256(original_payload)`.
```
jm-clius: `original_payload` needs a definition before being used here. Perhaps you can just change `"original message"` in the Terminology section to `"original payload"`?

standards/application/segmentation.md
```diff
@@ -0,0 +1,182 @@
...
+- Split the payload into one or more **data segments**, each of size up to `segmentSize` bytes.
+- Optionally generate **parity segments** using Reed–Solomon erasure coding, at the predefined parity rate.
```
jm-clius: It's a bit strange to have an "Optionally" as part of a MUST block. Perhaps you can change this paragraph to something like:

```
When the original payload exceeds...the sender:
- MUST compute...
- MUST split the payload...
- MAY use Reed-Solomon erasure coding...
```

We need a credible source that introduces Reed-Solomon to link to in order to make this spec self-sufficient (not introduce random terms). For example, what is the "predefined parity rate"? If we have a source, we can say something like "at a parity rate reasonable for the application (see [source])"

standards/application/segmentation.md
```diff
@@ -0,0 +1,182 @@
...
+
+When the original payload exceeds `segmentSize`, the sender **MUST**:
+
+- Compute a 32-byte `entire_message_hash = Keccak256(original_payload)`.
+- Split the payload into one or more **data segments**, each of size up to `segmentSize` bytes.
+- Optionally generate **parity segments** using Reed–Solomon erasure coding, at the predefined parity rate.  
+  Implementations **MUST NOT** produce more than 256 total segments (data + parity).
+- Encode each segment as a `SegmentMessageProto` with:
+  - The `entire_message_hash`
+  - Either data-segment indices (`segments_count`, `index`) or parity-segment indices (`parity_segments_count`, `parity_segment_index`)
```
jm-clius: At this point it's not clear what an "index" is (i.e. just a seq number for the segments). We could define inline, but another strategy is to introduce the Wire Protocol (`SegmentMessageProto`) first with a short explanation of the fields, followed by this section that describes how that protobuf is populated. This ordering helps a lot in making sure all terms and concepts are defined before being used in a heuristic.

standards/application/segmentation.md
```diff
@@ -0,0 +1,182 @@
...
+- Send all segments as individual Waku envelopes, preserving application-level metadata (e.g., content topic).
+
+Messages smaller than or equal to `segmentSize` **SHALL** be transmitted unmodified.
+
+### Receiving
+
+Upon receiving a segmented message, the receiver **MUST**:
+
+- Validate each segment according to [Wire Format → Validation](#validation).
+- Persist received segments, indexed by `entire_message_hash` and sender.
```
jm-clius: Would "buffer" or "cache" be a better word than "persist", which seems to imply some persistence medium

jm-clius (changes requested):
I have a few comments below, none of them major. Thanks!

jazzz (commented):
Overall, I think the idea has good bones.

message_hash, segment_index, segment_count provide base functionality for segmentation within an encrypted context.

Relaxing some of the implementation constraints would make it easier for other engineers to incorporate this work into their projects.

standards/application/segmentation.md
```diff
@@ -0,0 +1,182 @@
...
+title: Message Segmentation and Reconstruction over Waku
+name: Message Segmentation and Reconstruction
+tags: [waku-application, segmentation]
+version: 0.1
+status: draft
+---
+
+## Abstract
+
+This specification defines an application-layer protocol for **segmentation** and **reconstruction** of messages carried over Waku when the original payload exceeds the maximum Waku's message size. Applications partition the payload into multiple Waku envelopes and reconstruct the original on receipt, even when segments arrive out of order or up to a **predefined percentage** of segments are lost. The protocol uses **Reed–Solomon** erasure coding for fault tolerance. Messages whose payload size is **≤ `segmentSize`** are sent unmodified.
```
fryorcraken:
```suggestion
This specification defines an application-layer protocol for **segmentation** and **reconstruction** of messages carried over a message transport/delivery services with size limitation, when the original payload exceeds said limitation. Applications partition the payload into multiple wire-messages and reconstruct the original on receipt, even when segments arrive out of order or up to a **predefined percentage** of segments are lost. The protocol uses **Reed–Solomon** erasure coding for fault tolerance. Messages whose payload size is **≤ `segmentSize`** are sent unmodified.
```

standards/application/segmentation.md
```diff
@@ -0,0 +1,182 @@
...
+## Motivation
+
+Waku Relay deployments typically propagate envelopes up to **150 KB** as per [64/WAKU2-NETWORK - Message](https://rfc.vac.dev/waku/standards/core/64/network#message-size). To support larger application payloads, a segmentation layer is required. This specification enables larger messages by partitioning them into multiple envelopes and reconstructing them at the receiver. Erasure-coded parity segments provide resilience against partial loss or reordering.
+
+## Terminology
+
+- **original message**: the full application payload before segmentation.  
+- **data segment**: one of the partitioned chunks of the original message payload.  
+- **parity segment**: an erasure-coded segment derived from the set of data segments.  
+- **segment message**: a Waku envelope whose `payload` field carries a serialized `SegmentMessageProto`.
```
fryorcraken:
```suggestion
- **segment message**: a wire-message whose `payload` field carries a serialized `SegmentMessageProto`.  
```

fryorcraken (commented):
I think we can and should make the spec mostly Waku agnostic. Waku can still be mentioned in the motivation.

standards/application/segmentation.md
```diff
@@ -0,0 +1,182 @@
...
+- Validate each segment according to [Wire Format → Validation](#validation).
+- Persist received segments, indexed by `entire_message_hash` and sender.
+- Attempt reconstruction when the number of available (data + parity) segments equals or exceeds the data segment count.
+- Reconstruct by:
+  - Concatenating data segments if all are present, or
+  - Applying Reed–Solomon decoding if parity segments are available.
+- Verify `Keccak256(reconstructed_payload)` matches `entire_message_hash`.  
+  On mismatch, the message **MUST** be discarded and logged as invalid.
+- Once verified, the reconstructed payload **SHALL** be delivered to the application.
+- Incomplete reconstructions **SHOULD** be garbage-collected after a timeout.
```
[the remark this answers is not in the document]
plopezlpz: changed

standards/application/segmentation.md
```diff
@@ -0,0 +1,182 @@
...
+## Segmentation
+
+### Sending
+
+When the original payload exceeds `segmentSize`, the sender **MUST**:
+
+- Compute a 32-byte `entire_message_hash = Keccak256(original_payload)`.
+- Split the payload into one or more **data segments**, each of size up to `segmentSize` bytes.
+- Optionally generate **parity segments** using Reed–Solomon erasure coding, at the predefined parity rate.  
+  Implementations **MUST NOT** produce more than 256 total segments (data + parity).
```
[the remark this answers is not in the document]
plopezlpz: you are right, removing this, just picked from the library but in reality this is nor a hard limit

standards/application/segmentation.md
```diff
@@ -0,0 +1,213 @@
...
+Erasure-coded parity segments provide resilience against partial loss or reordering.
+
+## Terminology
+
+- **original payload**: the full application payload before segmentation.
+- **data segment**: one of the partitioned chunks of the original message payload.
+- **parity segment**: an erasure-coded segment derived from the set of data segments.
+- **segment message**: a wire-message whose `payload` field carries a serialized `SegmentMessageProto`.
+- **`segmentSize`**: configured maximum size in bytes of each data segment's `payload` chunk (before protobuf serialization).
+- **sender public key**: the origin identifier used for indexing persistence.
```
fryorcraken: plausible deniability flag cc @jazzz

standards/application/segmentation.md
```diff
@@ -0,0 +1,213 @@
...
+  uint32 parity_segment_index   = 5; // zero-based sequence number for parity segments
+  uint32 parity_segments_count  = 6; // number of parity segments (> 0)
+}
+```
+
+**Field descriptions:**
+
+- `entire_message_hash`: A 32-byte Keccak256 hash of the original complete payload, used to identify which segments belong together and verify reconstruction integrity.
+- `index`: Zero-based sequence number identifying this data segment's position (0, 1, 2, ..., segments_count - 1).
+- `segment_count`: Total number of data segments the original message was split into.
```
fryorcraken:
```suggestion
- `segments_count`: Total number of data segments the original message was split into.
```
are they 2 different things or the same?

standards/application/segmentation.md
```diff
@@ -0,0 +1,213 @@
...
+- **Intended for:** all Waku nodes at the application layer
+
+---
+
+## Security Considerations
+
+### Privacy
+
+`entire_message_hash` enables correlation of segments that belong to the same original message but does not reveal content.
+Traffic analysis may still identify segmented flows.
```
fryorcraken: Maybe add encryption considerations.

fryorcraken (approved):

standards/application/segmentation.md
```diff
@@ -0,0 +1,213 @@
...
+tags: [waku-application, segmentation]
+version: 0.1
+status: draft
+---
+
+## Abstract
+
+This specification defines an application-layer protocol for **segmentation** and **reconstruction** of messages carried over a message transport/delivery services with size limitation, when the original payload exceeds said limitation.
+Applications partition the payload into multiple wire-messages envelopes and reconstruct the original on receipt,
+even when segments arrive out of order or up to a **predefined percentage** of segments are lost.
```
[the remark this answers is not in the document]
jm-clius: Mmm, open for debate, but I still think it's most natural for SDS to apply after segmentation - otherwise SDS-R would also be a very large hammer, requesting repairs based on a message id that requires multiple broadcast chunks. It will be possible to work around it, but IMO SDS is best applied if the causality tree matches what is actually broadcast.

standards/application/segmentation.md
```diff
@@ -0,0 +1,213 @@
...
+```protobuf
+syntax = "proto3";
+
+message SegmentMessageProto {
+  // Keccak256(original payload), 32 bytes
+  bytes  entire_message_hash    = 1;
+
+  // Data segment indexing
+  uint32 index                  = 2; // zero-based sequence number; valid only if segments_count > 0
+  uint32 segment_count          = 3; // number of data segments (>= 2)
```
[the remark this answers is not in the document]
jazzz: ref: https://github.com/waku-org/specs/pull/91#discussion_r2449877856

standards/application/segmentation.md
```diff
@@ -0,0 +1,182 @@
...
+  uint32 parity_segments_count  = 6; // number of parity segments (> 0)
+}
+```
+
+### Validation
+
+Receivers **MUST** enforce:
+
+- `entire_message_hash.length == 32`
+- **Data segments:** `segments_count >= 2` **AND** `index < segments_count`
```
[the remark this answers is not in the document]
jazzz: Adding on here; Implementations like [ReliableChannels](https://forum.vac.dev/t/introducing-the-reliable-channel-api/580) or [Conversations](https://github.com/waku-org/specs/pull/93) may want to include the segmentation frame regardless to maintain a deterministic parsing tree.

If the output from the "Segmentation" step is sometimes a SegmentationMessageProto and sometimes a Application frame, this adds extra uncertainty as to how payloads ought to be decoded.

While deterministic parsing trees can be debated, it seems that forcing `segments_count > 1` is an optimization rather than a compatibility issue

jm-clius (approved):
Good job! I think this can be merged once the few outstanding comments are addressed.
In my mind they are:
- make the spec Waku-agnostic (can be mentioned in Motivation section)
- remove "sender public key" and just mention more generically that index can take place by sender as well as entire message hash.

standards/application/segmentation.md
```diff
@@ -0,0 +1,182 @@
+---
+title: Message Segmentation and Reconstruction over Waku
+name: Message Segmentation and Reconstruction
+tags: [waku-application, segmentation]
+version: 0.1
+status: draft
```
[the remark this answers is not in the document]
igor-sirotin: @Cofson I'm taking over this PR now. This spec was written from an existing implementation. I'm fine with both `raw` and `draft`, but maybe you have suggestion?

standards/application/segmentation.md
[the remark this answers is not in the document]
igor-sirotin: I wonder if we could work around the need for fetching segments, which are not needed (when enough segments have already been received). 

**Here's one idea I have:**
We could encode information into `messageId`, when calling SDS `wrapOutgoingMessage`. For exapmle:
1. Calculate/generate a message id X for the whole message
2. Segment it into N chunks
3. Pass chunk K into SDS, setting "sds message id" as `X-K:N`.

This way, from a single SDS message, the recipient will know all parameters: X, K and N. Hence, if enough chunks were received for X, then all missing SDS messages (we can reconstruct their IDs) can be marked to be "ignored":
    - don't invoke SDS-R or store nodes query for the missing chunks
    - (maybe) even mark the missing chunks as received, so that the sender doesn't retransmit them, though we don't know if other nodes received enough chunks, so this might be a bad idea.

---

I feel that this is a bit quirky and has more implications, but what do you think @jm-clius?

Also, I see some redundancy in using Reed-Solomon together with SDS(-R). Maybe when used with SDS, it makes sense to disable Reed-Solomon in segmentation? Then layers become clean, and there's no redundancy in retransmissions.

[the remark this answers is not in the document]
jm-clius: Yes, I think the problem with such an approach is that it brings higher-layer logic into the SDS Layer, which can confuse the state machine (and other assumptions) significantly. The way SDS works is to build a causality tree that can be traversed backwards (this message/chunk was preceded by this chunk that were in turn preceded by the one before, and so forth) or replayed (from past chunk X, the following chunks followed). If we now have higher-layer logic "inform" SDS that actually some of the causal dependencies are not really necessary, we break a host of assumptions (e.g. requesting specific missing chunks from another participant who has marked that chunk as "received", but meant to say "don't need it, as I can infer the message without it"). It may be possible to build something like this, but while all these protocols are in their infant stages it will make matters considerably more complex. Hence:

> Also, I see some redundancy in using Reed-Solomon together with SDS(-R). Maybe when used with SDS, it makes sense to disable Reed-Solomon in segmentation?

I would strongly advocate for this in our initial protocol stack. We need to harden the basics here in the cleanest/simplest stack we can imagine, and then proceed to add more sophisticated mechanisms at a later stage.

standards/application/segmentation.md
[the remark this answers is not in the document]
Cofson: @igor-sirotin sure thing - will you move it to lips-index or will it remain here for now? 

I would suggest to go with `raw` until you reach a review phase. If it's already there, `draft` also works.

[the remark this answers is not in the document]
igor-sirotin: > sure thing - will you move it to lips-index or will it remain here for now?

Yes, we're getting all open PRs merged now, and then these docs can be moved too.
Otherwise we'll lose the history of reviews, which is not ideal.

> I would suggest to go with raw until you reach a review phase. If it's already there, draft also works.

At the moment it's really about formally defining it, while we already have the working implementation in Status app. I'll keep it `draft` then and we can move on from there.

Thank you!

standards/application/segmentation.md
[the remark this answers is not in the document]
igor-sirotin: Alright, let's keep it as simple as possible for now and not try to oversee all cases.

Regarding Reed-Solomon, added a note to set `parityRate = 0` in Reliable Channel API https://github.com/logos-messaging/specs/pull/104, which brings Segmentation and SDS together.

Thanks @jm-clius

standards/application/segmentation.md
igor-sirotin: I was confused myself about `sender public key` too, and then noted your comments here.
Removed `sender public key` from this list now.

standards/application/segmentation.md
igor-sirotin: `segments_count` is the correct one, fixed

standards/application/segmentation.md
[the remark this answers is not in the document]
igor-sirotin: I believe the name of field should be plural `segments_count`, no matter if it's `>1` or `>=2`.
I'll fix it and close this comment, and resolve the minimum segments count value in this thread: https://github.com/waku-org/specs/pull/91#discussion_r2449877856.

standards/application/segmentation.md
```diff
@@ -0,0 +1,213 @@
...
+
+- **MUST** compute a 32-byte `entire_message_hash = Keccak256(original_payload)`.
+- **MUST** split the payload into one or more **data segments**,
+  each of size up to `segmentSize` bytes.
+- **MAY** use Reed–Solomon erasure coding at the predefined parity rate.
+- Encode each segment as a `SegmentMessageProto` with:
+  - The `entire_message_hash`
+  - Either data-segment indices (`segments_count`, `index`) or parity-segment indices (`parity_segments_count`, `parity_segment_index`)
+  - The raw payload data
+- Send all segments as individual Waku envelopes,
```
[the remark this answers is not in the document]
igor-sirotin: Agree. Fixed in fc44d154ebbfd3b142d82eb6488fc01d65f55fb8

standards/application/segmentation.md
```diff
@@ -0,0 +1,213 @@
...
+**Field descriptions:**
+
+- `entire_message_hash`: A 32-byte Keccak256 hash of the original complete payload, used to identify which segments belong together and verify reconstruction integrity.
+- `index`: Zero-based sequence number identifying this data segment's position (0, 1, 2, ..., segments_count - 1).
+- `segment_count`: Total number of data segments the original message was split into.
+- `payload`: The actual chunk of data or parity information for this segment.
+- `parity_segment_index`: Zero-based sequence number for parity segments.
+- `parity_segments_count`: Total number of parity segments generated.
+
+A message is either a **data segment** (when `segment_count > 0`) or a **parity segment** (when `segment_count == 0`).
```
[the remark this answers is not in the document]
igor-sirotin: I've cleaned up the menions of `segments_count`, I think it's clear enogh now

standards/application/segmentation.md
igor-sirotin: Fixed in bef0d205fcfbd2e52897e00a6daf4c40aad0189c

standards/application/segmentation.md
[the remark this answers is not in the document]
igor-sirotin: I agree that defaults should be `segments_count > 1`, i.e. always send `SegmentMessageProto`, even if the payload fits into a single transport message. And if decoding of transport message as `SegmentMessageProto` failed, it should be discarded.

On the other hand, it was introduced in Status and required some backwards compatibility. To achieve this, the messages that fit into a single transport message were send without `SegmentMessageProto` wrap.

To move one, yet keep Status' implementation legit, I've added a `Backwards Compatibility` section.
Full change in 47cf193df90add22f9f3789f0e196a5fedb785f3. @jazzz @jm-clius please let me know what you think.

https://github.com/logos-messaging/specs/blob/bef0d205fcfbd2e52897e00a6daf4c40aad0189c/standards/application/segmentation.md?plain=1#L122-L138

standards/application/segmentation.md
```diff
@@ -0,0 +1,239 @@
...
+
+## Backwards Compatibility
+
+Implementations **MAY** support a **backwards-compatible mode**, intended for deployments where this specification is being introduced incrementally and some peers do not yet implement segmentation.
+The mode is controlled by the `backwardsCompatible` configuration option, which defaults to `false`.
+
+When `backwardsCompatible = true`, the [Sending](#sending) procedure is amended as follows:
+
+- Payloads with size **≤ `segmentSize`** **SHALL** be transmitted unmodified, i.e., not wrapped in `SegmentMessageProto`.
+- Payloads exceeding `segmentSize` are wrapped and sent unchanged from [Sending](#sending).
```
osmaczko: If that’s the case, then there is no backward compatibility.

I think the Backwards Compatibility section should be removed altogether, as it doesn’t fit the segmentation spec. How breaking changes are introduced to the protocol should be up to the client. See: https://forum.research.logos.co/t/breaking-changes-and-roll-out-strategies/338#generic-roll-out-strategy-1

standards/application/segmentation.md
```diff
@@ -0,0 +1,239 @@
+---
+title: Message Segmentation and Reconstruction
+name: Message Segmentation and Reconstruction
+tags: [waku-application, segmentation]
```
jm-clius: I guess the `waku-application` tag is not relevant anymore :)

standards/application/segmentation.md
```diff
@@ -0,0 +1,239 @@
...
+version: 0.1
+status: draft
+---
+
+## Abstract
+
+This specification defines an application-layer protocol for **segmentation** and **reconstruction** of messages carried over a transport/delivery service with a message-size limitation, when the original payload exceeds said limitation.
+Applications partition the payload into multiple transport messages and reconstruct the original on receipt,
+even when segments arrive out of order or up to a **predefined percentage** of segments are lost.
+The protocol uses **Reed–Solomon** erasure coding for fault tolerance.
```
jm-clius:
```suggestion
The protocol optionally uses **Reed–Solomon** erasure coding for fault tolerance.
```

standards/application/segmentation.md
```diff
@@ -0,0 +1,239 @@
...
+
+A message is either a **data segment** (when `segments_count > 0`) or a **parity segment** (when `segments_count == 0`).
+
+### Validation
+
+Receivers **MUST** enforce:
+
+- `entire_message_hash.length == 32`
+- **Data segments:**
+  `segments_count >= 1` **AND** `index < segments_count`
```
jm-clius: Perhaps wise to also define a maximum `segments_count` (e.g. `256`) here, especially because receivers will allocate an in-memory cache based on expected number of segments.

standards/application/segmentation.md
```diff
@@ -0,0 +1,239 @@
...
+- `segmentSize` — maximum size in bytes of each data segment's payload chunk (before protobuf serialization).
+    **REQUIRED** parameter, configurable by the client.
+- `parityRate` — fraction of parity shards relative to data shards.
+    Configurable by the client. Defaults to **0.125** (12.5%).
+- `maxTotalSegments` — maximum number of total shards (data + parity) per message.
+    Implementation-specific parameter, fixed. The reference implementation uses **256**.
+
+**Reconstruction capability:**
+With the predefined parity rate,
+reconstruction is possible if **all data segments** are received or if **any combination of data + parity** totals at least `dataSegments` (i.e., up to the predefined percentage of loss tolerated).
```
jm-clius: Not very important now that we're unlikely to use RS, but as far as I can tell a receiver who happens to receive a bunch of parity segments (only) wouldn't know when to attempt reconstruction as these will all have `segments_count` set to `0`? If that is the case, having `segment_count=0` be an implicit indicator of a `parity` type is probably not a good idea, as you lose valuable information.

jm-clius (approved):
I think this is good to merge now! Thanks. I'd say future updates can be done in follow-up PRs. Minor comments below.

standards/application/segmentation.md
osmaczko: There must be at least one data segment to attempt reconstruction. Parity-only recovery would require having `len(paritySegments) >= len(dataSegments)`, which doesn't make much sense.

> having `segment_count=0` be an implicit indicator of a `parity` type is probably not a good idea

I agree, it is a bit confusing. The indicator could be `parity_segments_count` instead, then:

> A message is either a **data segment** (when `parity_segments_count == 0`) or a **parity segment** (when `parity_segments_count > 0`).

standards/application/segmentation.md
```diff
@@ -0,0 +1,239 @@
...
+
+## Wire Format
+
+Each segmented message is encoded as a `SegmentMessageProto` protobuf message:
+
+```protobuf
+syntax = "proto3";
+
+message SegmentMessageProto {
+  // Keccak256(original payload), 32 bytes
```
Ivansete-status, line 43: I think this is too implementation-specific. Is better to just mention it is a 32-byte array and leave it open for the app which algorithm to use.

standards/application/segmentation.md
```diff
@@ -0,0 +1,239 @@
...
+  bytes  entire_message_hash    = 1;
+
+  // Data segment indexing
+  uint32 index                  = 2; // zero-based sequence number; valid only if segments_count > 0
+  uint32 segments_count         = 3; // number of data segments
+
+  // Segment payload (data or parity shard)
```
Ivansete-status, line 50:
```suggestion
  // Segment payload (data or parity)
```

standards/application/segmentation.md
```diff
@@ -0,0 +1,239 @@
...
+
+### Receiving
+
+Upon receiving a segmented message, the receiver:
+
+- **MUST** validate each segment according to [Wire Format → Validation](#validation).
+- **MUST** cache received segments
+- **MUST** attempt reconstruction when the number of available (data + parity) segments equals or exceeds the data segment count:
+  - Concatenating data segments if all are present, or
+  - Applying Reed–Solomon decoding if parity segments are available.
```
Ivansete-status: It worth extending this point a bit more
```suggestion
- **MUST** attempt reconstruction when the total number of available segments (data + parity combined) equals or exceeds `segments_count`.
  `segments_count` is the **minimum number of segments** required to reconstruct the original payload: any combination of `segments_count` segments (data, parity, or mixed) is sufficient.
  Reconstruction proceeds by concatenating data segments if all are present, or by applying Reed–Solomon decoding if some data segments are missing but enough parity segments are available to reach the threshold.
```

standards/application/segmentation.md
Ivansete-status: I think we can re-phrase that one because the `dataSegments` is apparently not mentioned anywhere.
```suggestion
With the predefined parity rate, reconstruction is possible if any combination of data and parity segments totals at least `segments_count` (i.e., up to `parity_segments_count` segment losses are tolerated).
```

Ivansete-status: Yes I agree 100%
I think is better to have an explicit field for that.
Something like:
```nim
  // Segment type
  bool   is_parity              = 7; // true for parity segments, false (default) for data segments
```

Ivansete-status (commented):
Some nits so far

standards/application/segmentation.md
igor-sirotin: Discussion about it was initially started in https://github.com/logos-messaging/specs/pull/91#discussion_r2449877856, i.e. should `segmentation spec` force small messages to be wrapped in a `SegmentMessageProto`. And this is definitely a segmentation spec question, not application-level protocol. And my answer to this: yes, I believe it should. For simplicity and consistency.

Now, if we write it like that in the spec, then Status implementation becomes incompatible with this spec, and I wanted to avoid it. And changing it would involve a migration path, so we probably won't bother.

So I defined a generic way for applications to introduce segmentation with some Backwards Compatibility.

---

> If that’s the case, then there is no backward compatibility.

Yes, my suggestion covers most of the cases, except old client receiving segmented messages.
But this is a legit limitation. It's like introducing a new message type, which old clients will silently ignore.

_Could this be done just by the app itself?_
On reception, applications could attempt to decode both as `SegmentMessageProto` and application-level message. But this only enables new clients to receive messages from old clients, not the opposite.

So I do think this is a segmentation protocol-level matter.

standards/application/segmentation.md
igor-sirotin: @jm-clius it's defined here:
https://github.com/logos-messaging/specs/blob/9bcf528cfbe7b0399a29f5d64f773e6d1cdbd86b/standards/application/segmentation.md?plain=1#L165-L166

Or do you think we should introduce hard limits for implementations on spec level 🤔

standards/application/segmentation.md
igor-sirotin: > I think is better to have an explicit field for that.

I discussed this wtih @osmaczko yesterday. A separate field would be cleaner indeed, but we already have an implementation, which I'm trying to compatible with this spec. 

I'm not sure how hard this rule is though, maybe we should let it go and write the spec as we want it?
cc @osmaczko @jazzz @Ivansete-status

standards/application/segmentation.md
igor-sirotin: If we don't specify a specific hash, then different implementations will be incompatible with each other 🤔

standards/application/segmentation.md
igor-sirotin: `shard` here comes from Reed-Solomon coding. It's uses "data shards" and "parity shards" terminology.
it reads without it, but `shard` adds a bit of precision.

standards/application/segmentation.md
igor-sirotin: Agree, the structure was confusing, thanks!

I've used a bit different wording, let me know if this works for you:
https://github.com/logos-messaging/specs/blob/b7f9add1970ae8f4ca16feee33aeb2e6c81d4ebf/standards/application/segmentation.md?plain=1#L110-L112

standards/application/segmentation.md
jazzz: > Discussion about it was initially started in #91 (comment), i.e. should segmentation spec force small messages to be wrapped in a SegmentMessageProto. 

Point of clarity: That comment asks a different question - Should the spec force that there be a minimum of 2 segments. It seems that there is no harm in allowing a message to be segmented into a single segment.

standards/application/segmentation.md
```diff
@@ -0,0 +1,238 @@
...
+- Incomplete reconstructions **SHOULD** be garbage-collected after a timeout.
+
+---
+
+## Backwards Compatibility
+
+Implementations **MAY** support a **backwards-compatible mode**, intended for deployments where this specification is being introduced incrementally and some peers do not yet implement segmentation.
+The mode is controlled by the `backwardsCompatible` configuration option, which defaults to `false`.
+
+When `backwardsCompatible = true`, the [Sending](#sending) procedure is amended as follows:
```
jazzz: [Pebble] This boolean flag appears out of place, as it describes an actual implementation rather than protocol operation. 

From what I see `backwardsCompatible = true` does not result in a change to any of the payload types defined.

From what I understand - This section could be replaced with "Applications MAY choose not to follow this protocol and send arbitrary payloads" -  Which is always true.

standards/application/segmentation.md
jazzz: > Now, if we write it like that in the spec, then Status implementation becomes incompatible with this spec, and I wanted to avoid it. And changing it would involve a migration path, so we probably won't bother.

I'm not sure I follow the intention for this specification. Is it:
 - describing `SegmentMessageProto` and its valid fields?
 - or is describing an Application implementation?

jazzz: If the intention is to describe a protocol: I would agree with @osmaczko  that this section should be removed as it does not provide direction to implementors on when a payload can be considered valid. 

Examples:
CiCuA91ZDPecIslOkeZDvSxwXdCbG0KPxzNYq/AQgTneZxAAGAEiQC3GUqfnDm39JrE9kNtgB7yFSNcxpBzzdVbJN7KGF1vOdyy76UeeCeI1jM0fWK33PsRzJvtUpxSP0UyDOt5huA0oADAA


Protobuf uses schema dependent encoding, which is structurally ambiguous. This is an issue here because its possible for types generated by one schema to be decoded as another, without throwing an error. How are developers supposed to know if the given payload was generated with the SegmentMessageProto, or by a payload which was similar?

jazzz (approved):
Thank you for getting this across the line. 

The changes look good, and I think the core here is solid.

My only hang up is that the "Backwards compatibility" section currently has more questions than answers. I would highly recommend removing it for now, and then address it in a follow up.

However I won't block this PR.

standards/application/segmentation.md
jazzz: > maybe we should let it go and write the spec as we want it?

Its my belief that specifications ought to be more "aspirational" than "reflective of the current state". Especially if the current state is considered undesirable. 

> but we already have an implementation, which I'm trying to compatible with this spec.

Specifications are a tool to generate buy-in and promote inter-operability between implementors. The implementations should reflect that, not the other way around

igor-sirotin:
Here's what I'll change:
1. Forget about Status implementation. We're writing a new thing for ourselves.
2. Require wrapping single-segment messages (i.e. `segments_count >= 1`)
3. Remove `Backwards Compatibility` section
4. Add `is_parity` proto field instead of `segment_count=0` as implicit sign of parity shards. Adding 2 bytes is not harmful, but brings clarity.

standards/application/segmentation.md
igor-sirotin: Added an `is_parity` flag.

igor-sirotin:
I know some of you guys already approved, but I've resolved the comments and and made some important (but small) changes to the spec. Would appreciate if you can have another look.
cc @jm-clius @jazzz @osmaczko @Ivansete-status

standards/application/segmentation.md
```diff
@@ -0,0 +1,226 @@
...
+tags: [segmentation]
+version: 0.1
+status: draft
+---
+
+## Abstract
+
+This specification defines an application-layer protocol for **segmentation** and **reconstruction** of messages carried over a transport/delivery service with a message-size limitation, when the original payload exceeds said limitation.
+Applications partition the payload into multiple transport messages and reconstruct the original on receipt,
+even when segments arrive out of order or up to a **predefined percentage** of segments are lost.
```
Ivansete-status, line 13:
```suggestion
even when segments arrive out of order or up to a **predefined percentage** of segments are lost if Reed-Solomon is used.
```

standards/application/segmentation.md
```diff
@@ -0,0 +1,226 @@
...
+
+Receivers **MUST** enforce:
+
+- `entire_message_hash.length == 32`
+- `segments_count >= 1`
+- **Data segments** (`is_parity == false`):
+  `index < segments_count`
+- **Parity segments** (`is_parity == true`):
+  `parity_segments_count > 0` AND `parity_segment_index < parity_segments_count`
+
```
Ivansete-status, line 85: We may need to enforce max limit constants in the spec.
```suggestion
- **Data segments** (`is_parity == false`): segments_count < MaxAllowedNumSegments`
- **Data segments** (`is_parity == false`): sizeof(payload) < MaxAllowedPayloadSizeInBytes`
```

standards/application/segmentation.md
```diff
@@ -0,0 +1,226 @@
...
+message SegmentMessageProto {
+  // Keccak256(original payload), 32 bytes
+  bytes  entire_message_hash    = 1;
+
+  // Data segment indexing
+  uint32 index                  = 2; // zero-based sequence number for data segments
+  uint32 segments_count         = 3; // number of data segments (>= 1)
+
+  // Segment payload (data or parity shard)
+  bytes  payload                = 4;
```
Ivansete-status, line 51: I think is better to be a bit more explicit, to avoid confusion with original_payload
```suggestion
  bytes  segment_payload                = 4;
```

standards/application/segmentation.md
```diff
@@ -0,0 +1,226 @@
...
+
+## Segmentation
+
+### Sending
+
+To transmit a payload, the sender:
+
+- **MUST** compute a 32-byte `entire_message_hash = Keccak256(original_payload)`.
+- **MUST** split the payload into one or more **data segments**,
+  each of size up to `segmentSize` bytes.
```
Ivansete-status, line 97: I think we need to make sure the whole segment fits in the max allowed `segmentSize`. Not only the `segment_payload`.
```suggestion
- **MUST** split the payload into one or more **data segments**,
  making sure each `SegmentMessageProto` is up to `segmentSize` bytes.
```

standards/application/segmentation.md
```diff
@@ -0,0 +1,226 @@
...
+- **MAY** use Reed–Solomon erasure coding at the predefined parity rate.
+- **MUST** encode every segment as a `SegmentMessageProto` with:
+  - The `entire_message_hash`
+  - `segments_count` (total number of data segments, always set)
+  - When Reed–Solomon parity is used, `parity_segments_count` (total number of parity segments, set on every segment)
+  - For data segments: `is_parity = false`, `index`
+  - For parity segments: `is_parity = true`, `parity_segment_index`
+  - The raw payload data
+- Send each segment as an individual transport message according to the underlying transport protocol,
+  preserving application-level metadata (e.g., content topic).
```
Ivansete-status: Sorry I don't fully get this comment.

standards/application/segmentation.md
```diff
@@ -0,0 +1,226 @@
...
+
+### Integrity
+
+Implementations **MUST** verify the Keccak256 hash post-reconstruction and discard on mismatch.
+
+### Denial of Service
+
+To mitigate resource exhaustion:
+
+- Limit concurrent reconstructions and per-sender storage
```
Ivansete-status: I don't fully understand that one.

Ivansete-status (approved):
LGTM! Thanks for it! 🙌 
Adding some minor comments/questions.
Besides, I think is interesting to explicitly set the maximum recommended size per segment and the maximum recommended number of segments, which seems 256. Maybe interesting to have a `Recommendations section`.
I wonder if we should explicitly set endianness to be used to avoid any possible issue.
Cheers.

standards/application/segmentation.md
```diff
@@ -0,0 +1,226 @@
...
+- Enforce timeouts and size caps
+- Validate segment counts (≤ 256)
+- Consider rate-limiting at the transport layer (for example, via [17/WAKU2-RLN-RELAY](https://rfc.vac.dev/waku/standards/core/17/rln-relay) on Waku)
+
+### Compatibility
+
+Nodes that do **not** implement this specification cannot reconstruct large messages.
```
osmaczko:
```suggestion
Nodes that do **not** implement this specification cannot reconstruct any messages.
```

osmaczko (approved):

standards/application/segmentation.md
```diff
@@ -0,0 +1,226 @@
...
+
+```protobuf
+syntax = "proto3";
+
+message SegmentMessageProto {
+  // Keccak256(original payload), 32 bytes
+  bytes  entire_message_hash    = 1;
+
+  // Data segment indexing
+  uint32 index                  = 2; // zero-based sequence number for data segments
```
jazzz:
```suggestion
  uint32 index                  = 2; // zero-indexed sequence number for data segments
```

standards/application/segmentation.md
```diff
@@ -0,0 +1,226 @@
...
+To transmit a payload, the sender:
+
+- **MUST** compute a 32-byte `entire_message_hash = Keccak256(original_payload)`.
+- **MUST** split the payload into one or more **data segments**,
+  each of size up to `segmentSize` bytes.
+  A payload of size ≤ `segmentSize` produces a single data segment (`segments_count == 1`).
+- **MAY** use Reed–Solomon erasure coding at the predefined parity rate.
+- **MUST** encode every segment as a `SegmentMessageProto` with:
+  - The `entire_message_hash`
+  - `segments_count` (total number of data segments, always set)
```
jazzz: >  always set)

[Sand] In the future this spec could use some clarity on which feilds must be present and which are optional. Currently defined in prose which is harder to follow than a table or structured tagging

e.g `segments_count` [Required]   Total number of data segments

standards/application/segmentation.md
```diff
@@ -0,0 +1,226 @@
...
+- **MUST** cache received segments
+- **MUST** attempt reconstruction once at least `segments_count` distinct segments (data and parity combined) have been received:
+  - If all data segments are present, concatenate their `payload` fields in `index` order.
+  - Otherwise, recover the payload via Reed–Solomon decoding over the available data and parity segments.
+- **MUST** verify `Keccak256(reconstructed_payload)` matches `entire_message_hash`.
+  On mismatch,
+  the message **MUST** be discarded and logged as invalid.
+- Once verified,
+  the reconstructed payload **SHALL** be delivered to the application.
+- Incomplete reconstructions **SHOULD** be garbage-collected after a timeout.
```
jazzz: [Dust] This seems like an implementation suggest at best, and irrelevant at worst. What if my language doesn't have a garbage collector?

Software engineers are good at solving software engineering problems - focus on the parts they don't know.

standards/application/segmentation.md
```diff
@@ -0,0 +1,226 @@
...
+
+---
+
+## Implementation Suggestions
+
+### Reed–Solomon
+
+Implementations that apply parity **SHALL** use fixed-size shards of length `segmentSize`.
+The last data chunk **MUST** be padded to `segmentSize` for encoding.
+The reference implementation uses **nim-leopard** (Leopard-RS) with a maximum of **256 total shards**.
```
jazzz, line 134: [Pebble] Don't define protocol requirements in the Implementation suggestions section.

> The last data chunk **MUST** be padded to `segmentSize` for encoding.

This isn't a suggestion its a requirement.

standards/application/segmentation.md
```diff
@@ -0,0 +1,226 @@
...
+
+### Storage / Persistence
+
+Segments **MAY** be persisted (e.g., SQLite) and indexed by `entire_message_hash` and by sender. Sender MAY be authenticated, this is out of scope of this spec.
```
jazzz: [Dust] The use of *MAY* here seems not in line with  2119. See Section 6.

"""For example, they must not be used to try to impose a particular method
   on implementors where the method is not required for
   interoperability."""

standards/application/segmentation.md
```diff
@@ -0,0 +1,226 @@
...
+Implementations **SHOULD** support:
+
+- Duplicate detection and idempotent saves
+- Completion flags to prevent duplicate processing
+- Timeout-based cleanup of incomplete reconstructions
+- Per-sender quotas for stored bytes and concurrent reconstructions
```
jazzz, line 144: [Dust] As an implementor I'm not getting much value from this section.

The implementations suggestions section is a place to help fellow developers by. guiding them towards good solutions. Talking about the issues, and how to avoid them would have more value for me personally.

" As segments arrive over time, the memory footprint of the cache will grow. To reduce this, its recommended that older segments be cleared out to reduce the memory footprint. Using a fixed size  circular buffer or running a cleanup process to remove older segments would address this."

standards/application/segmentation.md
jazzz: [Sand] This does not appear to be a security considerations

standards/application/segmentation.md
```diff
@@ -0,0 +1,226 @@
...
+---
+
+## Implementation Suggestions
+
+### Reed–Solomon
+
+Implementations that apply parity **SHALL** use fixed-size shards of length `segmentSize`.
+The last data chunk **MUST** be padded to `segmentSize` for encoding.
+The reference implementation uses **nim-leopard** (Leopard-RS) with a maximum of **256 total shards**.
+
```
jazzz, line 135: [Dust] I'd love to hear some guidance on when/which circumstances to use parity segments and when they lead to excess bandwidth utilization

standards/application/segmentation.md
jazzz: > I think this is too implementation-specific.

I don't see a blocker with requiring Keccak256 here. Its simple enough, and provides clarity.

> ... then different implementations will be incompatible with each other

Is the expectation that all implementations of this spec can interop? 

Seems like interoperability would be limited to each "Application" as the lower layers and the upper layers are not defined. 

I imagine that application interoperability could only be achieved through a integration spec that defines how all the layers could be combined, and what parameterizations to use.

standards/application/segmentation.md
```diff
@@ -0,0 +1,226 @@
...
+- **MUST** validate each segment according to [Wire Format → Validation](#validation).
+- **MUST** cache received segments
+- **MUST** attempt reconstruction once at least `segments_count` distinct segments (data and parity combined) have been received:
+  - If all data segments are present, concatenate their `payload` fields in `index` order.
+  - Otherwise, recover the payload via Reed–Solomon decoding over the available data and parity segments.
+- **MUST** verify `Keccak256(reconstructed_payload)` matches `entire_message_hash`.
+  On mismatch,
+  the message **MUST** be discarded and logged as invalid.
+- Once verified,
+  the reconstructed payload **SHALL** be delivered to the application.
```
jazzz, line 125: [Dust] This section focuses on the receiver role, and reads more like an Implementation spec than a Protocol spec. It works. But I think it could be tighter and clearer if it focused on normative declarations about the wireformats/payloads/types. At its core this specification is describing a wireformat, involving people and roles distracts from the details. 


In general: 
- Define the abstract types that this spec is generic over: (e.g ReassembledPayloadBytes, InputPayloadBytes)
- Define the concrete types which are introduce in this protocol (e.g SegmentMessageProto)
- define the requirements of creating and converting these types.

By making it about the payloads, you automatically remove all the noise. 


E.g
- SegmentMessageProto.segments_count MUST be >= 1
- SegmentMessageProto.is_parity MAY be omitted or set to 'false'; If so all properties must be ignored.
- If SegmentMessageProto.is_parity is `true`  `SegmentMessageProto.parity_segments_count` must be > 1


For ReassembledPayloadBytes to be valid:

- Every SegmentMessageProto's used to compose  ReassembledPayloadBytes MUST have unique `index` values;

- ReassembledPayloadBytes MUST be composed by concatenating SegmentMessageProto.payloads in order of SegmentMessageProto.index

- Every SegmentMessageProto's used to compose  ReassembledPayloadBytes MUST have a `entire_message_hash` value equal to keccak256(ReassembledPayloadBytes);


Invalid `ReassembledPayloadBytes` must be discarded

standards/application/segmentation.md
```diff
@@ -0,0 +1,226 @@
...
+
+  // Segment type
+  bool   is_parity              = 7; // true for parity segments, false (default) for data segments
+}
+```
+
+**Field descriptions:**
+
+- `entire_message_hash`: A 32-byte Keccak256 hash of the original complete payload, used to identify which segments belong together and verify reconstruction integrity.
+- `index`: Zero-based sequence number identifying this data segment's position (0, 1, 2, ..., segments_count - 1). Set only on data segments.
```
jazzz: [Dust] There is an unneeded asymmetry here.

`segments_count` -> `parity_segments_count`
`index` -> `parity_segment_index`

Consider normalizing the field names.

standards/application/segmentation.md
```diff
@@ -0,0 +1,226 @@
...
+- `segments_count`: Total number of data segments the original message was split into. Set on every segment (data and parity).
+- `payload`: The actual chunk of data or parity information for this segment.
+- `parity_segment_index`: Zero-based sequence number for parity segments. Set only on parity segments.
+- `parity_segments_count`: Total number of parity segments generated. Set on every segment (data and parity) when Reed–Solomon parity is used; `0` (default) otherwise.
```
jazzz: [Dust] Find replace `segments_` -> `segment_`

Plurals are just going to lead to spelling mistakes and inconsistencies. Using a common roots will be easier to maintain.

standards/application/segmentation.md
```diff
@@ -0,0 +1,226 @@
...
+
+**API simplicity:**
+Libraries **SHOULD** require only `segmentSize` from the application for normal operation.
+
+### Support
+
+- **Language / Package:** Nim;
+  **Nimble** package manager
+- **Intended for:** application-layer use over any transport with message-size constraints
+
```
jazzz: [??!?] I'm not entirely sure how to interpret this

Is the author suggesting that the protocol ought to only support nim?
Is the author sharing a personal opinion of their favorite language?

I don't understand why this is is here; or how it impacts a implementor at all.

jazzz (approved):
I think if I had to go implement this, my library would be compatible with others.  However I also believe there is much room for improvement and simplification. 

- This spec has a single payload with 7 fields (all basic types); It takes 66 lines to describe them. Thats not unreasonable, but room from improvement. 

- I think this work could benefit from following other examples such as: [66/Metadata](https://github.com/logos-co/logos-lips/blob/master/docs/messaging/standards/core/66/metadata.md). Ruthless, efficient, clear.

- As a developer I wish the details were displayed cleaner especially given how small the surface area is. In future updates it would be helpful to reduce the amount of repetition across sections. e.g.
   - Field descriptions:  `segments_count: Total number of data segments the original message was split into. Set on every segment (data and parity).`
   - Validation: `Receivers **MUST** enforce: segments_count >= 1`
    - Receiving: `the receiver: **MUST** encode every segment as a SegmentMessageProto with: segments_count (total number of data segments, always set)`

standards/application/segmentation.md
igor-sirotin: I think such details are not needed for Abstract

standards/application/segmentation.md
igor-sirotin: @jazzz I wanted to argue that "segments count" is the correct way of saying.
Then I remembered who of us knows English better and checked with AI. Apparently I was saying it wrong all my life...

standards/application/segmentation.md
igor-sirotin: Renamed to `data_segment_count` and `data_segment_index`

igor-sirotin: Also I realised that we don't need both `data_segment_idnex` and `parity_segment_index`. Will address in a separate PR.

standards/application/segmentation.md
```diff
@@ -0,0 +1,182 @@
...
+No other combinations are permitted.
+
+---
+
+## Implementation
+
+### Reed–Solomon
+
+Implementations that apply parity **SHALL** use fixed-size shards of length `segmentSize`.  
+The last data chunk **MUST** be padded to `segmentSize` for encoding.
```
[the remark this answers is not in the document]
igor-sirotin: Padding is a requirement of Reed-Solomon, all segments must be of the same size

standards/application/segmentation.md
igor-sirotin: Hmmm, I think it's pretty obvious that `SegmentMessageProto` contains a sinle segment payload, not the original payload

standards/application/segmentation.md
igor-sirotin: I've added the maxTotalSegments validation, because if it's over the limit, Reed-Solomon won't be able to reconstruct the message.

But I think we can be soft about payload size. This validation happens on reception, and the limit is coming from transport service. If the segment already made it through the limited transport, why drop it, if we can still parse it.

standards/application/segmentation.md
igor-sirotin: I agree, but this will be not as simple as changing this line. Let me do it in a separate PR.

standards/application/segmentation.md
igor-sirotin: I've rewritten it like this:
https://github.com/logos-messaging/specs/blob/745d1c2be5e28636638616c7ca465a40ddf8d0ed/standards/application/segmentation.md?plain=1#L193-L194

This is about preventing possible attacks where receiver would have to process and hold in memory many segments until they're reconstructed, e.g.:
- receiving many concurrent messages
- receiving many messages with the same `entire_message_hash`
- receiving little messages, but with large payloads

standards/application/segmentation.md
igor-sirotin: Removed it, as it's already mentioned here: https://github.com/logos-messaging/specs/blob/745d1c2be5e28636638616c7ca465a40ddf8d0ed/standards/application/segmentation.md?plain=1#L143

standards/application/segmentation.md
igor-sirotin: Moved here: https://github.com/logos-messaging/specs/blob/dd3de0b211f9b7e36050a36e3822e3779c197001/standards/application/segmentation.md?plain=1#L99

Also, padding is not needed when Reed-Solomon is not enabled, as we are simply concatenating segments.

standards/application/segmentation.md
igor-sirotin: Indeed. Moved under `Deployment Considerations`

standards/application/segmentation.md
igor-sirotin: Removed this entire section, makes no sense here 💯

standards/application/segmentation.md
igor-sirotin: Removed this part

igor-sirotin:
I'm gonna merge now.

My initial goal was to address the comments in this PR, as it was ready and mostly approved, so I didn't bother much. I agree there're more problems to the current spec. I will address them in separate PRs, after moving it to https://github.com/logos-co/logos-lips.

PR #89, opened 2025-10-17
introduce reliable channel API

fryorcraken:
The "Reliable Channel" API is the layer that fits between "Chat SDK" and "Waku API", it provided e2e reliability via SDS, as well as other useful utilities for developers.

It fits in more refined proposed layers for Waku: https://github.com/waku-org/pm/pull/364

It is partially implemented in js-waku, see https://github.com/waku-org/js-waku/pull/2683 for the most up to date version (blocked on some [`subscribe`](https://github.com/waku-org/specs/pull/86) and [`send`](https://github.com/waku-org/specs/pull/87) implementations for the Waku API).

See forum post (TODO) for more details.

PR #93, opened 2025-10-26
Chat Private 

jazzz:
This PR adds a specification which describes a conversation protocol for pairwise communication called PRIVATE1. 

It covers payload generation and operation after a key has been agreed upon by both parties.

### Notes 
This may eventually be simplified and use [Reliable Channels](https://github.com/waku-org/specs/pull/89) but its explicit in how it utilizes SDS + segmentation and the reasons for doing so.


standards/application/privatev1.md
```diff
@@ -0,0 +1,417 @@
...
+### Message Reliability
+
+Scalable Data Sync (SDS) is used to detect missing messages, provide delivery confirmation, and handle retransmission of payloads.
+SDS is implemented according to the [specification](https://github.com/vacp2p/rfc-index/blob/main/vac/raw/sds.md).
+
+**SDS Field Mappings**
+
+The following mappings connect PRIVATE1 concepts to SDS fields:
+
+- `sender_id`: !TODO: This requires PRIVATE1 to be identity aware
```
jazzz: Unsure exactly what the best course of action is here. Recent changes to SDS requires an sender_id be attached the the SdsMessage. 

This then brings in a concept of identity to this specification. Its definitely possible to introduce Identity concepts in order to pass them through to SDS, however it seems like a lot of added complexity

standards/application/privatev1.md
```diff
@@ -0,0 +1,417 @@
...
+
+**Sender Validation**
+SDS uses a `sender_id` payload field to determine whether a message was sent by the remote party. This value is sender reported and not validated which can have unknown implications if trusted in other contexts. For security hygiene Clients SHOULD drop SDS messages if `sender_id` != the sender derived from the encryption layer. !TODO: PRIVATE1 is not sender aware currently
+
+**Bloom Filter Configuration**
+
+PRIVATE1 uses bloom filter parameters of `n=2000` (expected elements) and `p=0.001` (false positive probability).
+This configuration produces bloom filters of approximately 3.5 KiB per message.
+
+!TODO: Can the bloom filter be dropped in 1:1 communication?
```
kaichaosun: I agree with the split.
Only use casual history is enough for message sent check.

standards/application/privatev1.md
```diff
@@ -0,0 +1,417 @@
...
+## Copyright
+
+Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).
+
+
+## References
+- **[DOUBLERATCHET]** "The Double Ratchet Algorithm", Signal, 2016.
+  https://signal.org/docs/specifications/doubleratchet/
+
+- **[SDS]** "Scalable Data Sync Specification", vac, 2024.
```
jonesmarvin8: All of these github references from vacp2p and waku-org are broken.

standards/application/privatev1.md
```diff
@@ -0,0 +1,417 @@
...
+By segmenting first, the reliability layer can track individual segments and request retransmission of only the missing fragments.
+
+**Implementation**
+
+The segmentation strategy used is defined by [!TODO: Flatten link once completed](https://github.com/waku-org/specs/pull/91)
+
+Implementation specifics:
+- Error correction is not used, as reliable delivery is already provided by lower layers. 
+- `segmentSize` = `max_seg_size`
+- All payloads regardless of size are wrapped in a segmentation message.
```
osmaczko: This contradicts segmentation specs:
> Messages whose payload size is ≤ segmentSize are sent unmodified.

osmaczko (approved):
Looks good overall. The reasoning behind the `Content -> Segmentation -> Reliability -> Encryption` layering makes sense.

Please remember to address the TODOs before merging.

standards/application/privatev1.md
```diff
@@ -0,0 +1,417 @@
...
+tags:
+editor: Jazz Alyxzander (@Jazzz)
+contributors:
+---
+
+# Abstract
+
+This specification defines PRIVATE1, a conversation protocol for establishing secure, full-duplex encrypted communication channels between two participants. PRIVATE1 provides end-to-end encryption with forward secrecy and post-compromise security using the Double Ratchet algorithm, combined with reliable message delivery via Scalable Data Sync (SDS) and efficient segmentation for transport-constrained environments.
+
+The protocol is transport-agnostic and designed to support both direct messaging and as a foundation for group communication systems. PRIVATE1 ensures payload confidentiality, content integrity, sender privacy, and message reliability while remaining resilient to network disruptions and message reordering.
```
igor-sirotin, line 14: I'm not sure but something sounds wrong in `... support both direct messaging and as a foundation ...`. I get the meaning, just not sure of the language.

I think you know better, I'm just highlighting
```suggestion
The protocol is transport-agnostic and designed to support direct messaging and serve as a foundation for group communication systems. PRIVATE1 ensures payload confidentiality, content integrity, sender privacy, and message reliability while remaining resilient to network disruptions and message reordering.
```

standards/application/privatev1.md
```diff
@@ -0,0 +1,417 @@
...
+- **Content Integrity**: Recipients can detect if message contents were modified by a third party. Any tampering with encrypted payloads will cause decryption to fail, preventing corrupted messages from being accepted as authentic.
+- **Sender Privacy**: Only the recipient can determine who the sender was. Observers cannot identify the sender from encrypted payloads, though both participants can authenticate each other's messages.
+- **Forward Secrecy**: A compromise in the future does not allow previous messages to be decrypted by a third party. Message keys are deleted immediately after use and cannot be reconstructed from current state, even if long-term keys are later compromised.
+- **Post-Compromise Security**: Conversations eventually recover from a key compromise. After an attacker loses access to a device, the security properties are eventually restored.
+- **Dropped Message Observability**: Messages lost in transit are eventually observable to both sender and recipient. 
+
+## Definitions
+
+This document makes use of the shared terminology defined in the [CHAT-DEFINITIONS](https://github.com/waku-org/specs/blob/jazzz/chatdefs/informational/chatdefs.md) specification.
```
igor-sirotin:
```suggestion
This document makes use of the shared terminology defined in the [CHAT-DEFINITIONS](informational/chatdefs.md) specification.
```

standards/application/privatev1.md
```diff
@@ -0,0 +1,417 @@
...
+Content MUST be smaller than `255 * max_seg_size`
+due to segmentation protocol limitations.
+
+**Structure**
+
+The protocol treats the contents as an arbitrary sequence of bytes and is agnostic to its contents.
+
+### Payload Delivery
+How payloads are sent and received by clients is deliberately not specified by this protocol.
+Transport choice is an implementation decision that should be made based on deployment requirements.
```
igor-sirotin: Am I reading this right? Or does it actually mean "PrivateV1 implementation"?
```suggestion
Transport choice is an application decision that should be made based on deployment requirements.
```

standards/application/privatev1.md
```diff
@@ -0,0 +1,417 @@
...
+- `conversation_id` - globally unique identifier
+
+To maintain the security properties:
+- `sk` MUST be known only by the participants.
+- `sk` MUST be derived in a way that ensures mutual authentication of the participants
+- `sk` SHOULD have forward secrecy by incorporating ephemeral key material 
+- `rsk` and `ssk` SHOULD incorporate ephemeral key material
+
+PRIVATE1 requires a unique identifier, however the exact derivation is left to implementations to determine.
+- `conversation_id` MUST be unique across all instances of chat conversations
```
igor-sirotin, line 104: What is meant by `chat` here? And what is `chat conversations`?

standards/application/privatev1.md
```diff
@@ -0,0 +1,417 @@
...
+## Architecture
+
+This conversation type assumes there is some service or application which wishes to generate and receive end-to-end encrypted content. 
+It also assumes that some other component is responsible for delivering the generated payloads. At its core this protocol takes the content provided and creates a series of payloads to be sent to the recipient.
+
+```mermaid
+flowchart LR
+    Content:::plain--> PrivateV1 --> Payload:::plain
+    classDef plain fill:none,stroke:transparent;
+```
```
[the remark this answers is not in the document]
igor-sirotin: I agree that Frames seem to be unnecessary here.

Maybe something like this?
```mermaid
flowchart LR
    Content:::plain--> PrivateV1 --> Payloads@{ shape: stacked-rectangle }
    classDef plain fill:none;
    class Payloads plain;
```

standards/application/privatev1.md
```diff
@@ -0,0 +1,417 @@
...
+- `sk` MUST be derived in a way that ensures mutual authentication of the participants
+- `sk` SHOULD have forward secrecy by incorporating ephemeral key material 
+- `rsk` and `ssk` SHOULD incorporate ephemeral key material
+
+PRIVATE1 requires a unique identifier, however the exact derivation is left to implementations to determine.
+- `conversation_id` MUST be unique across all instances of chat conversations
+- `conversation_id` SHOULD be consistent across applications to maintain interoperability
+
+Additionally implementations MUST determine the following constants:
+- `max_seg_size` - maximum segmentation size to be used.
```
igor-sirotin, line 108: Isn't this dictated by the transport used by an application? 
If yes, then this `max_seg_size` should be defined by application, not protocol implementation? Or am I not getting "implementation" term here?

igor-sirotin, line 108: If `segmentSize` is exactly `max_seg_size`, should we mention that actual payloads can and will be over larger than this limit?

standards/application/privatev1.md
```diff
@@ -0,0 +1,417 @@
...
+All other frame types are protocol-owned and intended for client processing, not application consumption.
+
+This establishes a critical invariant: any frame that is not `content` is meant for the protocol layer.
+When a client encounters an unknown frame type, it can definitively conclude this represents a version compatibility issue.
+
+**Processing Rules**
+
+- All application-level content MUST use the `content` frame type
+- Clients SHALL only pass `content` frames to applications
+- Clients MAY drop unrecognized frame types
```
igor-sirotin, line 249: Is "clients" same thing as "implementations"?
```suggestion
- Implementations SHALL only pass `content` frames to applications
- Implementations MAY drop unrecognized frame types
```

igor-sirotin (approved):
I wish I read it earlier!

standards/application/privatev1.md
jazzz: >  Isn't this dictated by the transport used by an application?

When this was modeled it implemented relaibility inside of the converastion. 

> then this max_seg_size should be defined by application, n

sure. However we're trying to avoid a circular dependency. `max_seg_size` is a paramter that must be passed in.

jazzz: No. Thats covered at a higher level. the `max_seg_size` doesn't have to match the MTU of the transport layer. Its just a parameter. 

The WakuChat protocol spec would force these constraints.

PR #97, opened 2025-11-26
Ratcheting Private Identifiers

jazzz:
This PR adds a draft spec for Ratcheting Private Identifiers (RPIs). RPIs are an extension to DH ratchet based encryption systems that provides a mechanism for privacy preserving payload discovery and conversation binding.

Outstanding Work:
- [x] Upgrade to HKDF-Extract
- [ ] Unify Sender and Receiver chains
- [ ] Investigate removing `chainSecret` it appears unnecessary
- [ ] Technical Proof (link: )
- [ ] Replace diagrams with text embedded versions


standards/application/rpi.md
```diff
@@ -0,0 +1,195 @@
...
+---
+
+## Abstract
+
+This specification defines Ratcheting Private Identifiers (RPIs), a cryptographic mechanism that enables efficient payload discovery in end-to-end encrypted systems without leaking metadata. RPIs allow recipients to identify which messages are intended for them while maintaining sender anonymity and conversation unlinkability. The mechanism uses key ratcheting to generate single-use identifiers that are transmitted in cleartext but remain meaningless to observers. RPIs provide forward privacy, post-unmasking privacy, and payload unlinkability properties. This specification is designed to integrate with existing encryption protocols such as Double Ratchet, providing a privacy-preserving solution to the key selection problem in multi-sender encrypted messaging systems.
+
+## Background
+
+End-to-end encrypted chat protocols like Double Ratchet provide strong message privacy and forward secrecy.
+However, they don't address a fundamental question: how does a recipient know which key to use for decryption?
```
kaichaosun: Would be good to know how other products solve this problem.

standards/application/rpi.md
```diff
@@ -0,0 +1,195 @@
...
+
+The problem is similar to receiving an encrypted postcard.
+With a single sender, you know who sent it.
+With multiple senders, it's unclear whose keys were used during encryption.
```
kaichaosun: Is it ok to just hash the key or identifier?

standards/application/rpi.md
```diff
@@ -0,0 +1,195 @@
...
+Ratcheting private identifiers (RPIs) are a means for receiving clients to associate a given payload to an encryption state in a privacy-preserving manner.
+RPIs are single-use identifiers transmitted unencrypted alongside encrypted payloads.
+The privacy guarantees come from the fact that only conversation participants can derive which RPIs to expect, making the identifiers meaningless to observers.
+
+Both clients maintain a `PI` and `NextPI` for both sending and receiving directions.
+By performing a symmetric ratchet step with secret key material (`chainSecret`), clients derive the next `PI` in a chain.
+Each ratchet step updates `PI` and derives an RPI by truncating the updated value.
+
+The system integrates with an external encryption mechanism that provides `RK` and `chainSecret`.
+When a DH ratchet step occurs, `NextPI` replaces `PI` and a new `NextPI` is derived from the updated `RK`.
```
kaichaosun: Would be good to know more about `RK` here and when `RK` gets updated?

kaichaosun (commented):
The spec makes sense to me, but I do feel it's complex for metadata protection. 
I have following 2 ideas, which is not as private as this spec, but worth to have a discussion: 
1. since message key is already using double ratchet, we can use the hash of message key to help with decryption process. 
2. I'm wondering if we are going to rotate in the identify design, if so, I think the hash of identify could be used in the decryption process.

jazzz:
Closing for now. The problem is changing as the system migrates to PostQuantum encryption. Will reintroduce at a later date

